PR #313, opened 2019-01-11
Line 54 - self:EmitSound

Dodobro:
A sound will play on the exterior when someone tries to open a locked TARDIS in an attempt to add more immersion than just some text!


MattJeanes:
Thanks for this, I'll check it out in-game when I'm back from work ☺

MattJeanes:
Thanks! Appreciate your contribution.

PR #315, opened 2019-02-20
Possible Sound Fix?

Dodobro:
In the past people have complained that the sound will cut out if you switch from internal to external view mid-teleport. This method stops that and from various different tests it doesn't seem to break anything.


Win7yes:
Doesn't this play the sounds twice and with no 3D space?

Dodobro:
> Doesn't this play the sounds twice and with no 3D space?

It plays it once, however, the 3D sound is disabled.

kennyrkun:
3d sound disabled meaning the client hears it wherever they are?

I think that's a bad idea.

Dodobro:
> 3d sound disabled meaning the client hears it wherever they are?
> 
> I think that's a bad idea.

Dodobro:
We are looking into another alternative using https://wiki.garrysmod.com/page/Global/CreateSound

Issue #330, opened 2020-05-14
Move server-side settings into a UI panel


Win7yes:
People seem to have trouble with the concept of using the Source console so we'll just make things easier for them. Granted this should have been a thing from the start maybe, but I thought the console was fine for starters.

MattJeanes:
You will need to consider permissions as I believe currently there are no serverwide admin controls in the settings panel

Win7yes:
I could use the same toolpanel we use for the Interface button. I've asked around and people would rather have it separate from the screens

MattJeanes:
Yep that's a good idea, and they only are enabled for admins (ply:IsAdmin() or ply:IsSuperAdmin())

PR #331, opened 2020-05-14
Day 2 patch

Win7yes:
- Added setting for damage in menu (same panel where the interface button is)
- Disabled fire damaging the TARDIS (would cause damage from very far away for some unknown reason)
- Added prevention for going over the max health value


Win7yes:
This should close #330 by the way

lua/tardis/sh_menuoptions.lua
```diff
@@ -0,0 +1,42 @@
...
+					net.WriteType(val)
+				net.SendToServer()
+			end
+			panel:AddItem(htoggle)
+		end)
+	end)
+else
+	util.AddNetworkString("TARDIS-SvCvars")
+
+	net.Receive("TARDIS-SvCvars", function(len, ply)
```
MattJeanes, line 33: This allows arbitrary execution of commands, please limit it to just tardis2_damage and send up just the new value as an int don't send up the command too

MattJeanes:
Nice one I'll push this out tomorrow morning

PR #351, opened 2021-02-03
Allow traces to route through portals properly and improve custom portals support

MattJeanes:
- See Doors changes also included from https://github.com/MattJeanes/Doors/pull/11
- Scanner now hides the interior in the sky properly
- Custom portals can now be linked to a part and they switch on and off with them to improve performance and stop accidental teleports

### DIFF
--- a/addon.json
+++ b/addon.json
@@ -8,6 +8,7 @@
 		"*.txt",
 		"*.git*",
 		"*.fgd",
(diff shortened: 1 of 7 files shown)


francodml (dismissed):
All of this is looking great, can't find any objections.

francodml (approved):

PR #381, opened 2021-03-03
Physlock improvements

### LINKED ISSUE #362: Add a key for physlock in third person
It's usually convenient to lock the TARDIS at one place before you return to the real controls from third person.
I suggest G.

parar020100:
1. Physlock is now dependent on power and cannot be enabled if power is off
2. Added a key for physlock in 3rd person, "G" is the default
3. Added a bit of shaking in the interior when physlock is enabled
4. Instant physlock at any moment was unrealistic and op, so now depending on the speed:
  a. If speed is low (<1000), physlock engages normally
  b. If speed is medium (1000 - 2500), a visual explosion happens, but doesn't damage the TARDIS
  c. In case of fast flying, the explosion will be bigger and damage the TARDIS (should never be more than 15%) 
5. Fixed physlock message logic for Legacy TARDIS
6. Also, moved flight key into "Third Person" section, as discussed with @MattJeanes 

For (4), I added explosion strength. Default one is still 60, but if the TARDIS dies, it becomes bigger (300).

closes #362


lua/entities/gmod_tardis/modules/sh_health.lua
```diff
@@ -272,7 +273,7 @@ if SERVER then
 			self:SetData("prevortex-flight", false)
 			self:Mat()
 		end
-		self:Explode()
+		self:Explode(300)
```
MattJeanes, line 276: What is the reason for increasing this from 60 to 300?

lua/entities/gmod_tardis/modules/sh_physlock.lua
```diff
@@ -0,0 +1,88 @@
...
+		end
+	end,
+	key=KEY_G,
+	serveronly=true,
+	exterior=true
+})
+
+if SERVER then
+	function ENT:SetPhyslock(on)
+
```
MattJeanes, line 21: Remove empty line

lua/tardis/parts/legacy/legacy_physbrake.lua
```diff
@@ -11,8 +11,8 @@ PART.Sound = "tardis/control_handbrake.wav"
 
 if SERVER then
 	function PART:Use(ply)
-		local result = self.exterior:TogglePhyslock() or false
-		ply:ChatPrint("Physics Lock ".. (result and "engaged" or "disengaged"))
+		self.exterior:TogglePhyslock()
+		ply:ChatPrint("Physics Lock ".. (self.exterior:GetData("physlock") and "engaged" or "disengaged"))
```
MattJeanes, line 15: This print should be dependent on the above TogglePhyslock returning true, maybe put a message in the else like "failed to toggle physlock"?

AmyJeanes (changes requested):
Looking at the code it looks pretty good to me, couple of notes I've added but overall nice. Will give an in-game test once code review is finished

lua/entities/gmod_tardis/modules/sh_health.lua
parar020100: This is the death scenario. Shouldn't death explosion be bigger than normal explosion due to technical difficulties?

Also, it would stop people from just shooting others' TARDISes up close. 
Intentionally or not, legacy had this feature that if you exploded the TARDIS up close, it backfired. This also works now here. It's not major, you can still shoot it from a distance.

lua/tardis/parts/legacy/legacy_physbrake.lua
parar020100: TogglePhyslock() used to return current physlock state
Now it returns whether it was toggled successfully, as ToggleFloat() does
Basically, I didn't change anything - this is refactoring, not functionality change

Plus, this all will go away soon with centralized log/message system

lua/entities/gmod_tardis/modules/sh_physlock.lua
parar020100, line 21:
```suggestion
	function ENT:SetPhyslock(on)
```

lua/entities/gmod_tardis/modules/sh_physlock.lua
```diff
@@ -0,0 +1,88 @@
...
+		if not on and self:CallHook("CanTurnOffPhyslock") == false then
+			return false
+		end
```
parar020100, line 24:
```suggestion
	function ENT:SetPhyslock(on)
		if not on and self:CallHook("CanTurnOffPhyslock") == false then
			return false
		end
```

lua/entities/gmod_tardis/modules/sh_physlock.lua
parar020100: Done.
Also, btw, do you know you can suggest changes by yourself with this button:
изображение

MattJeanes:
Looks to be working but the screen shake on the interior seems a bit excessive especially since it always happens regardless of speed, maybe disable this and reimplement the legacy feature where damage would cause screen shake on the interior scaled to the damage amount?

AmyJeanes (approved):
Happy with the new lowered screen shake, approved!

PR #386, opened 2021-03-13
Interior tips

parar020100:
Added interior tips
- Tips are specified in **Tips** list in interior options, similar to **Parts**
- There are configurable positions, text, visibility, view range, colors and fonts
- The default tip style is similar to legacy one
Garry's Mod 13 03 2021 16_45_20
- Each interior can have its own tip style, specified in "default" tip (which isn't displayed by itself)
- This is the default tip style I made for legacy:
Garry's Mod 13 03 2021 16_46_54
- Each tip can be customized seperately, you only have to specify the options you want different from the default one, others are inherited
- There's a "learning mode" setting that enables and disables the tips. Reload or respawn are not required for it to change. Tips are on by default.
- Legacy TARDIS currently has all tips enabled. I wasn't sure about some names, so you can modify them. I've made most of those like legacy ones. Also, I'm not sure about "flight modes" which don't exactly exist in rewrite. It's up to you.


MattJeanes (approved):
I've made some tweaks but otherwise looking nice!

PR #387, opened 2021-03-13
Screen improvements

parar020100:
1. Legacy screen has been returned, since it can now be small with visual GUI added. 

2. You can now turn interior screens on and off, but there are actually three modes:
    a. Screen on, you can see the GUI
    gm_flatgrass0019
    b. Screen off, GUI is transparent, you can see everything that's behind it 
    (it has been requested by addon creators many times, as there are animations at screen parts)
    gm_flatgrass0020
    c. Power off, the screen is black
    gm_flatgrass0022

3. Interior screens can be toggled by real controls, as well as a button in the visual GUI (which can always be seen in the popup).
    
4. A button which toggles screens has been added to the legacy interior. 
    The model is same as "dematerialisation cirquit" one, but it is hidden within the screen
    изображение

5. Rainbow color-changing background has been removed. 
    Background is either transparent (GUI off) or black (GUI on or power off)

6. Interior can have ScreensEnabled option, which sets whether the Visual GUI in 3D screens is enabled by default.
    Without the option, they are enabled by default

7. I've renamed visual_gui_... and visualgui_... settings and elements to universal visgui_... 
    Some of the names were just too long
    
8. Each 3D screen in Interior.Screens{} can now have visgui_rows option, which specifies the default number of rows for it.
    This is useful for small screens. Even if the value is changed in the settings, the number of rows is still set to the interior one.
    
9. There's a visgui_rows_override setting. If on, interior visgui_rows option is ignored, and the number of rows is selected only by user.


AmyJeanes (changes requested):
Looking pretty nice, I've made some tweaks notably around switching stuff to use GetPower() and using hooks rather than hardcoding power state / screen toggle as being the defining factors of whether or not the screen is on/black. There is just one comment about the new visual GUI button to turn off the screens having no icon.

AmyJeanes (approved):

PR #391, opened 2021-03-15
Replace default

parar020100:
Replacing default interior with legacy, renaming all legacy stuff into default
+ Changed icons for legacy and TARDIS
+ Moved main entity from "Doctor Who" section into "Doctor Who - TARDIS" and made it first on the list
+ Added anti-gravs (float) control to default-legacy instead of "unused"


lua/entities/gmod_tardis/shared.lua
```diff
@@ -2,8 +2,8 @@
 
 ENT.Base="gmod_door_exterior"
 ENT.Spawnable=true
-ENT.PrintName="TARDIS Rewrite"
-ENT.Category="Doctor Who"
+ENT.PrintName=" TARDIS " -- spaces are for it to go first in the list
```
MattJeanes, line 5: Can we undo this change and the line below, when the time comes we will rename legacy to TARDIS Legacy but they should still be in the Doctor Who category

MattJeanes:
```suggestion
ENT.PrintName="TARDIS"
```

lua/entities/gmod_tardis/shared.lua
```diff
@@ -2,8 +2,8 @@
...
+ENT.Category="Doctor Who - TARDIS"
```
MattJeanes, line 6:
```suggestion
ENT.Category="Doctor Who"
```

lua/tardis/interiors/default.lua
```diff
@@ -1,124 +1,124 @@
--- Default
+-- Default (used to be Legacy)
```
MattJeanes, line 1: This wording is not necessary

```suggestion
-- Default
```

MattJeanes:
In-game testing passed with rewrite not installed, all working good

MattJeanes (changes requested):

MattJeanes (approved):

PR #396, opened 2021-03-15
TARDIS2 E2 extension compatibility for Legacy

francodml:
This replaces the original Wiremod E2 extension included in legacy with the new one intended for Rewrite, where function handling is offloaded to the entity.
Along with this, all E2 functions are now in a dispatch table inside `lua/entities/sent_tardis/init.lua`


francodml:
This is pending testing. I need to check I haven't broken anything.

lua/entities/gmod_wire_expression2/core/custom/tardis2.lua
```diff
@@ -0,0 +1,207 @@
+E2Lib.RegisterExtension("tardis2", true)
```
MattJeanes, line 1: Should probably just be "tardis" same with filename

lua/entities/sent_tardis/init.lua
```diff
@@ -1625,6 +1625,418 @@ function ENT:TogglePower()
 	return false
 end
 
+local function CheckPP(ply, ent) // Prop Protection
```
MattJeanes, line 1628: Use lua style comment --

lua/entities/sent_tardis/init.lua
```diff
@@ -1625,6 +1625,418 @@ function ENT:TogglePower()
...
+	return hook.Call("PhysgunPickup", GAMEMODE, ply, ent)
+end
+
+local E2Commands = {
+	["Demat"] = function(ent,data,pos,ang)
+		if ent and IsValid(ent) and CheckPP(data.player,ent) then
+			if not (ent:GetClass()=="sent_tardis") then return 0 end
```
MattJeanes, line 1635: You can get rid of these sent_tardis checks, maybe rename ent to self as well

lua/entities/sent_tardis/init.lua
```diff
@@ -1625,6 +1625,418 @@ function ENT:TogglePower()
...
+			else
+				return 0
+			end
+		end
+		return 0
+	end
+}
+
+function ENT:HandleE2(cmd, ...)
+	if not E2Commands[cmd] then return end
```
MattJeanes, line 2036: Should probably hard error here if not valid `error("TARDIS Handler for E2 command " .. cmd .. " is not registered")` or something

MattJeanes (changes requested):
Nice stuff, just a couple of little tweaks to do

lua/entities/sent_tardis/init.lua
francodml: Copied this from the old module without noticing, whoops

lua/entities/gmod_wire_expression2/core/custom/tardis.lua
```diff
@@ -1,575 +1,207 @@
...
-			return 1
-		else
-			return 0
-		end
+local function HandleE2(ent, name, ...)
+	if IsValid(getTardis(ent)) then
+		return ent:HandleE2(name, ...)
 	else
-		return 0
+		error("Can't call TARDIS functions on something other than a TARDIS.")
```
MattJeanes, line 23: Can you make sure that this doesn't crash wiremod if this happens? Should just return an E2 error

lua/entities/sent_tardis/init.lua
```diff
@@ -1625,6 +1625,388 @@ function ENT:TogglePower()
 	return false
 end
 
+local function CheckPP(ply, ent) -- Prop Protection
+	return hook.Call("PhysgunPickup", GAMEMODE, ply, ent)
+end
+
+local E2Commands = {
+	["Demat"] = function(self,data,pos,ang)
+		if self and IsValid(self) and CheckPP(data.player,self) then
```
MattJeanes, line 1634: You can remove the if self and IsValid(self) from all of these (keep the CheckPP bit though on ones that have that)

MattJeanes (changes requested):

lua/entities/gmod_wire_expression2/core/custom/tardis.lua
francodml: It doesn't crash wiremod, only the expression. I'm not sure how to return E2 errors

lua/entities/sent_tardis/init.lua
```diff
@@ -2003,7 +2003,10 @@ local E2Commands = {
 }
 
 function ENT:HandleE2(cmd, ...)
-	if not E2Commands[cmd] then return end
+	if not E2Commands[cmd] then 
+		error("TARDIS E2 Handler for command '"..cmd.."' is not registered")
```
MattJeanes, line 2007: This should probably have the ,0 thing as well for the E2 errors

lua/entities/gmod_wire_expression2/core/custom/tardis.lua
francodml: Done now, adding a zero as the level made the error be thrown as an e2 error, but it doesn't show lines and stuff.

lua/entities/sent_tardis/init.lua
francodml: I was going to counter that there's no way to manually call commands, but since we might implement things on the Rewrite that legacy won't have, it'll seem like our fault.

MattJeanes (approved):

PR #397, opened 2021-03-15
Multiple fixes and improvements

parar020100:
- Fixed setting name "Toggle" -> "Toggle Flight"
- Option for screen not to have black rectangle if power off
- Removed "Rewrite" from all texts
- Removed file for old screensaver
- Preset texts for tips
- Added destination control to default interior's keyboard


lua/entities/gmod_tardis_interior/modules/cl_tips.lua
```diff
@@ -34,6 +55,9 @@ function ENT:InitializeTips(style_name)
 		for setting,value in pairs(interior_tip) do
 			tip[setting]=value
 		end
+		if tip.control then
+			tip.text = tip_control_texts[tip.control] or ""
```
MattJeanes, line 59: the fallback should be clear it doesn't exist, should be like "unknown control: ".. tip.control

lua/entities/gmod_tardis_interior/modules/sh_screens.lua
```diff
@@ -39,12 +39,13 @@ ENT:AddHook("Initialize", "screens", function(self)
...
-				ext=self.exterior,
-				int=self,
-				visgui_rows=v.visgui_rows
+			self.screens3D[k] = TARDIS:LoadScreen(k, {
+				width = v.width,
+				height = v.height,
+				ext = self.exterior,
+				int = self,
+				visgui_rows = v.visgui_rows,
+				power_off_black = v.power_off_black,
```
MattJeanes, line 48: This should default to true (unless you disagree)?

lua/tardis/interiors/default.lua
```diff
@@ -42,11 +42,12 @@ T.Interior={
...
-			ang=Angle(0, 84.5, 91),
-			width=227.75,
-			height=140,
-			visgui_rows=2
+			pos = Vector(44.5, -6.5, 39),
+			ang = Angle(0, 84.5, 91),
+			width = 227.75,
+			height = 140,
+			visgui_rows = 2,
+			power_off_black = true
```
MattJeanes, line 50: This can be removed once/if it is default

MattJeanes (changes requested):
Some nice tweaks, finally dropping the rewrite naming(!)

lua/entities/gmod_tardis_interior/modules/sh_screens.lua
parar020100: It is true by default

lua/tardis/interiors/default.lua
parar020100: Yes, it can
However, it might be ok to have it here as literally everyone uses legacy as an example

lua/entities/gmod_tardis_interior/modules/sh_screens.lua
MattJeanes: Are you sure? It doesn't look like it only when explicitly set by the screen

lua/tardis/interiors/default.lua
MattJeanes: That's fair, okay can leave it in here then

lua/entities/gmod_tardis_interior/modules/cl_tips.lua
parar020100: Added an error

lua/entities/gmod_tardis_interior/modules/sh_screens.lua
parar020100: You're right, sorry, I was thinking about another thing. Fixed now.

MattJeanes (approved):

PR #401, opened 2021-03-15
Wiremod and E2 support

### LINKED ISSUE #253: Wiremod Integration
One big part of the fun with the legacy tardis was the compatbility with wiremod. Even if it was simply being able to input a vector and dematerialize.

I look forward to hearing back. Addon looks great so far!

Thanks,
John

francodml:
Adds Wiremod and E2 support to the TARDIS, as well as means for developers to handle their own functions and add their own Wire inputs and outputs.

E2 function names are consistent with Legacy, which means full backwards compatibility with Expressions meant for Legacy (tried and tested)
Unimplemented features such as cloaking just fail silently. Since the original Legacy functions were made to return a success state when called, this isn't a problem.

This essentially closes #253.

Changelog:

- Added Wiremod support
- Added E2 support: function names remain, all your favourite expressions will work with the new TARDIS.
- Added Demat, Mat, and Destination inputs, as well as a Health output.
- Bunch of internal stuff which helps us help you have a better TARDIS experience by making Wiremod friendlier to work with in code.

~A developer write-up will come up shortly, detailing the technical side of the Expression 2 and Wiremod interfaces.~
# Dev stuff
I've added a library to easily work with Wiremod.

Inputs and outputs can be easily added with

`self:AddWireInput(name, description, type)`

`self:AddWireOutput(name, description, type)`

These add Wire inputs and outputs. Name is the name displayed on the Wire tool, description is displayed alongside the name, in parenthesis, like "Name(Description)", type is the input/output type. 

Valid types are defined by WireLib [here](https://github.com/wiremod/wire/blob/438c9e8649eaac1bd4ddfad2fa2c4c4491290b0e/lua/wire/server/wirelib.lua#L69-L105 "lua/wire/server/wirelib.lua lines 69 to 105 @ github.com/wiremod/wire"), or you can print the WireLib.DT global table in your game. From my testing, a full install of Wiremod + Wire extras returns the WireLib.DT table below.

```lua
WireLib.DT = 	{
		ANGLE      = WireLib.DT.ANGLE,
		ANY        = WireLib.DT.ANY,
		ARRAY      = WireLib.DT.ARRAY,
		BIDIRARRAY = WireLib.DT.BIDIRARRAY,
		BIDIRTABLE = WireLib.DT.BIDIRTABLE,
		BONE       = WireLib.DT.BONE,
		COLOR      = WireLib.DT.COLOR,
		COMPLEX    = WireLib.DT.COMPLEX,
		EFFECT     = WireLib.DT.EFFECT,
		ENTITY     = WireLib.DT.ENTITY,
		GTABLE     = WireLib.DT.GTABLE,
		MATRIX     = WireLib.DT.MATRIX,
		MATRIX2    = WireLib.DT.MATRIX2,
		MATRIX4    = WireLib.DT.MATRIX4,
		NORMAL     = WireLib.DT.NORMAL,
		QUATERNION = WireLib.DT.QUATERNION,
		RANGER     = WireLib.DT.RANGER,
		STRING     = WireLib.DT.STRING,
		TABLE      = WireLib.DT.TABLE,
		TRACEDATA  = WireLib.DT.TRACEDATA,
		VECTOR     = WireLib.DT.VECTOR,
		VECTOR2    = WireLib.DT.VECTOR2,
		VECTOR4    = WireLib.DT.VECTOR4,
		WIRELINK   = WireLib.DT.WIRELINK
	}
```
As far as I'm aware, you can skip the description and type parameters, but never name, as that's how you're going to handle your wire input or output. Skipping type will set the type to ANY.

Handling inputs and outputs is done via the `self:TriggerWireOutput(name, value)` function and the `OnWireInput(name, value)` hook.

OnWireInput works like a standard hook. Name is the input name as defined via AddWireInput.
It's triggered whenever an input changes or is connected. I recommend checking that `value` is equal to something (ie, 1) before doing anything.

`self:TriggerWireOutput(name, value)` is a simple function. When called on the interior or exterior, it sets or triggers a Wire output. Name is the output's name, and value is any value the output can take.

E2 functions on the other hand are defined on an E2 module placed in entities/gmod_wire_expression2/core/custom. You can look at our E2 module to see how it works.

Basically, we route all E2 functions into a "handler". This was done not only for extensibility, but also to allow for backwards compatibility with Legacy. Whenever the handler function is called, the hook `HandleE2(self, name, expression, ...)` gets called. The name is the "command name" passed to ENT:HandleE2(). You can hook into the `HandleE2` hook on interior and exterior modules (via `ENT:AddHook()`). You can hook into existing E2 functions if you so wish, just check the E2 module to see their command names.


MattJeanes (approved):
Nice work!

PR #412, opened 2021-03-17
Debug messages

parar020100:
Includes a convar for debug

### DIFF
--- a/lua/tardis/libraries/sh_messages.lua
+++ b/lua/tardis/libraries/sh_messages.lua
@@ -1,0 +1,28 @@
+CreateConVar("tardis2_debug", 0, {FCVAR_ARCHIVE, FCVAR_REPLICATED}, "TARDIS - debug enabled")
+
+function TARDIS:IsDebugOn()
+	return GetConVar("tardis2_debug"):GetBool()
+end
+
+function TARDIS:Debug(text)
+	if TARDIS:IsDebugOn() then
+		local fulltext = "[TARDIS Debug]"..text
+		if CLIENT and LocalPlayer() then
+			LocalPlayer():ChatPrint(fulltext)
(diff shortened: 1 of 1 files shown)


lua/tardis/libraries/sh_messages.lua
```diff
@@ -0,0 +1,28 @@
+CreateConVar("tardis2_debug", 0, {FCVAR_ARCHIVE, FCVAR_REPLICATED}, "TARDIS - debug enabled")
+
+function TARDIS:IsDebugOn()
+	return GetConVar("tardis2_debug"):GetBool()
+end
+
+function TARDIS:Debug(text)
+	if TARDIS:IsDebugOn() then
+		local fulltext = "[TARDIS Debug]"..text
```
MattJeanes, line 9: Should probably be a space here before the message
```suggestion
		local fulltext = "[TARDIS Debug] "..text
```

lua/tardis/libraries/sh_messages.lua
```diff
@@ -0,0 +1,28 @@
+CreateConVar("tardis2_debug", 0, {FCVAR_ARCHIVE, FCVAR_REPLICATED}, "TARDIS - debug enabled")
+
+function TARDIS:IsDebugOn()
+	return GetConVar("tardis2_debug"):GetBool()
```
MattJeanes, line 4: This should be cached as TARDIS:Debug code may be running multiple times per frame and we do not want to call this every single time as it is relatively expensive, it should be retrieved once to set TARDIS.debug = true/false and updated at runtime using https://wiki.facepunch.com/gmod/cvars.AddChangeCallback

MattJeanes (changes requested):

PR #423, opened 2021-03-19
GUI themes and improvements 

parar020100:
- Made new structure for GUI themes
  - Themes are configured in lua files 
  - Themes can inherit each other.
  - Subolders and filenames are now confiigurable, structure is up to creator
  - Animated textures are now supported
  - Split frames from icons
    - Added frame types configurable for buttons
    - Frame type is specified by 2 numbers; default and clicked versions 
    - 0 = normal (blue), 1 = on (green), 2 = off (red)
- Replaced default theme with a better one, with HD and better aligned (symmetrical) frames
- Added new "Transparent" theme, which is also cool
- For buttons without icons (with text only), centred the text and made it completely readable.
  - Re-added text only theme, which doesn't look bad any more
- Made possible to specify button order in the hexagonal layouts
  - Renamed "Console" back to "Virtual Console"
  - Added button order for menu and virtual console
- Fixed titlebar's button offsets so that they never intersect
- Made "Destination" button open popup from 3D screen 
- Renamed "Destination" screen button to "Coordinates" without renaming the actual screen
- Moved third person and manual destination select into main menu from the virtual console
- Added icons for "TARDIS Status" custom screen
- Fixed bug which allowed to press invisible buttons near screen sides
- Fixed bug of pressing one button while trying to press another (by increasing offset between the buttons)
- Fixed bug: arrows appeared on settings screen after pressing "back"


parar020100:
P.S. CheckGuiTheme() is a blank function on purpose, will change soon, but now let's keep it this way so that there's no need to redo the block

lua/tardis/sh_visgui_themes.lua
```diff
@@ -0,0 +1,93 @@
+TARDIS.visgui_themes={}
+
+local theme_basefolder = "materials/vgui/tardis-themes/"
+
+local function CheckGUITheme(theme)
+	return true
```
MattJeanes: What's the purpose of this bit?

MattJeanes (approved):
I've made a few changes of myself you can check out and fixed a few issues introduced in this PR but otherwise looks great, nice work!

PR #426, opened 2021-03-20
Redecoration enhancements, spawnmenu context options, sequence conditions

francodml:
## Sequence conditions
Sequences can now define an "evaluator" function. If this function returns false explicitly, the sequence can't run.

This is also reflected on the highlighted control tips.

## Spawnmenu icons context menu
TARDIS icons now have a context menu. So far it allows you to set your preferred interior, as well as the interior used for redecoration.

## Redecoration enhancements
Mostly internal, redecor now uses TARDIS:SpawnTARDIS instead of manually creating the entity. Along with this the conditions allowing for redecoration have been tweaked, and a hook has been added (``ShouldRedecorate``)


MattJeanes (approved):
Great stuff!

PR #430, opened 2021-03-22
Fix #424 (Damage in vortex)

### LINKED ISSUE #424: TARDIS in vortex can interact with outside world
When in vortex, TARDIS can still interact** with the outside world. For example, you can place dynamite at a place where it was when it dematerialised, and kill the TARDIS in vortex. 
Vortex is spawned at the spot where TARDIS dematerialised. The easiest way to solve the issue is change that, I suggest we spawn the vortex at the top of the map, using an algorythm we have for interior spawning.

francodml:
This pull request focuses solely on fixing #424.
As such, the TARDIS can no longer incur damage whilst in the time vortex.

A list of changes is detailed below

- TARDIS can no longer take damage while in the vortex
- Fixed #424

### DIFF
--- a/lua/entities/gmod_tardis/modules/sh_health.lua
+++ b/lua/entities/gmod_tardis/modules/sh_health.lua
@@ -297,7 +297,7 @@ if SERVER then
 	end)
 
 	ENT:AddHook("ShouldTakeDamage", "Health", function(self, dmginfo)
-		if not TARDIS:GetSetting("health-enabled") or self:GetData("vortex",false) then return false end
+		if not TARDIS:GetSetting("health-enabled") then return false end
 	end)
 
 	ENT:AddHook("OnTakeDamage", "Health", function(self, dmginfo)
--- a/lua/entities/gmod_tardis/modules/sh_health.lua
+++ b/lua/entities/gmod_tardis/modules/sh_health.lua
@@ -296,8 +296,8 @@ if SERVER then
 		end
 	end)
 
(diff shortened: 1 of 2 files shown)


lua/entities/gmod_tardis/modules/sh_health.lua
```diff
@@ -296,8 +296,8 @@ if SERVER then
 		end
 	end)
 
-	ENT:AddHook("ShouldTakeDamage", "DamageOff", function(self, dmginfo)
-		if not TARDIS:GetSetting("health-enabled") then return false end
+	ENT:AddHook("ShouldTakeDamage", "Health", function(self, dmginfo)
+		if not TARDIS:GetSetting("health-enabled") or self:GetData("vortex",false) then return false end
```
MattJeanes, line 300: This should instead be a new hook in the teleport module, not in here

AmyJeanes (changes requested):

parar020100:
We need to make sure that interior still can take damage though

MattJeanes:
Good call @parar020100, perhaps dmginfo contains the entity that got damaged and we can use this? Combine that with moving this into a new hook in the teleport module and should be golden

francodml:
> @francodml
> We need to make sure that interior still can take damage though

Interior should still be able to, the interior and exterior hooks are separate. I'll check though

francodml:
I've moved the hook over to the teleport module. Interior can still take damage as previously thought, since the hooks are called separate from each other, in each entity's ENT:OnTakeDamage

PR #449, opened 2021-03-23
Killfeed Icon & Name

Dodobro:
Not sure why this hasn't been done already. Simply added an icon and name for the TARDIS when it kills a player.

image


lua/tardis/cl_killfeed.lua
```diff
@@ -0,0 +1,4 @@
+if CLIENT then
+    killicon.Add( "gmod_tardis", "killfeed/gmod_tardis", Color( 255, 255, 255, 255 ))
+    language.Add("gmod_tardis", "TARDIS")
+end
```
francodml, line 4: Since this is a `cl_` file it gets loaded on the client alone, so we don't need to check.
```suggestion
killicon.Add( "gmod_tardis", "killfeed/gmod_tardis", Color( 255, 255, 255, 255 ))
language.Add("gmod_tardis", "TARDIS")
```

francodml (changes requested):
Seems good, only one small change but nothing big. Will wait on Matt for this though.

lua/tardis/cl_killfeed.lua
Dodobro: I was always told that the prefix to a filename was purely for sorting on the developer's end, and didn't impact things in-game. Now I know, thanks Win!

francodml: In general, `cl_init.lua` `init.lua` and `shared.lua` in entities are actually loaded the way they're labelled (client, server, and shared/both respectively). This same principle works with the loader for the `lua/tardis` folder, files prefixed with `cl_` `sv_` or `sh_` get loaded on the specified realms only (or both for sh_).

MattJeanes: Yeah can we make this change please? Otherwise looks great!

MattJeanes (changes requested):
Looks mostly good, nice little tweak! Can we apply the fix @francodml suggested to remove the if CLIENT bit?

francodml (approved):
Nice, this passes for me.

MattJeanes (approved):

Issue #456, opened 2021-03-25
Fast flight with manual destination select


parar020100:
When you're choosing destination manually, you can still fly with a rather slow speed. It makes travelling experience boring - it's faster to noclip to the destination than to use the TARDIS.

We need to add a different speed for destination select, it should be possible to choose faster. I suggest we add exponential acceleration if SHIFT is pressed, and current speed if it isn't. Then, we get both speed and precision.

Issue #403, opened 2021-03-16
Power/Self-Repair toggle mid vortex flight


CrethFox:
It is possible to toggle power and self repair during vortex flight. 
This issue does not occur with the original legacy interior.

francodml:
Seems to have been fixed in #470, will close when that rolls out

Issue #368, opened 2021-03-02
Random location travel


parar020100:
This would be useable in many other issues, including HADS and advanced travel
I suggest a new control for it:
* We randomly check different points of the map to spawn the interior
* For a random destination, we do the same, but instead of choosing the highest available, we do "get to floor" for every found location ("get to floor" like R in destination select)
* Also, if TARDIS has travelled, it could randomly select out of previous locations

Brundoob:
You could create an internal table of some sort to backlog the last 10-20 locations so it can pick one from random (ignoring duplications) and pick randomly between the previous destinations or the random destination as mentioned above

Issue #357, opened 2021-02-22
Add proper Time Rotor Handbrake control


parar020100:
"2005 Improved" and "2010 Improved" have different versions of it
The best way to do it is to take best from both of them

Current implementations and ideas:
- [both] When handbrake on, TARDIS cannot dematerialise
- [2005] Special sound and effects for being unable to dematerialise (currently - shaking in 2005)
- [2010] A way to dematerialise with handbrake on, but explosions and damage ("engine release lever")
- [2010] It should prevent fast return and even HADS
- [both] Handbrake inside the vortex could lead to crash landing and major damage
- [2005] It should be left possible to block the ability to dematerialise by other circumstances (2005's artron absorber)
- [none] Maybe there should be a way to block materialisation too (for example, by time distortions, like "Angels take Manhatten" one), it could be easy to implement with this feature

Questions:
- Should it allow real world flight? In 2005 it does, in 2010 it used to be same as physlock (which it definetely shouldn't be, as it's often required to demat with physlock on)

CrethFox:
The artron absorber works for both demat and remat within a radius of the device
I dunno why thats split into two points if the absorber is going to be one

JEREDEK:
> The artron absorber works for both demat and remat within a radius of the device
> I dunno why thats split into two points if the absorber is going to be one

Actually, it doesn't fully work.
I have no ide why it doesn't work on release, since it worked on testing, but oh well.

Mrleplayz:
The reason i made the handbrake act as a physics lock is to force players to use forced demat if they absolutely need the handbrake on. It adds a reason to use forced takeoff. I'm not implying that i want it to stay that way, all i'm saying is i thought about the idea gameplay-wise.

parar020100:
> The reason i made the handbrake act as a physics lock is to force players to use forced demat if they absolutely need the handbrake on. It adds a reason to use forced takeoff. I'm not implying that i want it to stay that way, all i'm saying is i thought about the idea gameplay-wise.

Maybe in advanced mode, physlock would need to be on. But I think that making them synonims isn't very great

parar020100:
> 
> 
> The artron absorber works for both demat and remat within a radius of the device
> I dunno why thats split into two points if the absorber is going to be one

Because forbidding to materialise would also be useable if there's an ability to materialise in someone else's TARDIS
It would be as part of shields

francodml:
[...]

PR #470, opened 2021-03-31
Handbrake, failed mat/demat effects and many other changes

parar020100:
Includes: #469, it should be merged first
Related issues: #357 #456 #446 #368 #337 #403

- Handbrake control, prevents TARDIS from taking off
  - Sounds for failed demat
  - HADS makes an exception
  - Handbrake while demating cancells teleport with damage, sparks, explosions, power off
    - A setting to remove those effects for weak computers
    - Power can't be turned back on in 10 seconds after crash landing
  - Handbrake at vortex or while materialising results in **same**, but also with falling out at a random location in the map  
- Made impossible to demat with doors open
  - HADS makes an exception
  - Added a setting to disable that
- Engine release, allows demat with opened doors or handbrake with broken effects (shaking, explosions, sparks)
  - Only works if pressed right after failed demat (several seconds)
  - A setting to remove those effects for weak computers  
- Destination select improvements
  - Increased destination speed boost with "shift" 
  - "H" key in third person opens destination select
  - **Random location select in coordinates menu** (2 options - just random and random on the ground)
- Sounds for damaged TARDIS teleportation
- Fixed poweroff being possible in the vortex 
- Disabled cloister bells if power is disabled
- Fixed message for float in third person
- [internal] Fixed removing parts in interior overrides (Parts = { ... <part> = false, ... }


MattJeanes:
Haven't reviewed the code yet but playing around with it is really fun! Loving the new sounds and effects! I got a little confused at first why it wouldn't take off with the doors open, I think perhaps we should either add a message prompt or make it close them by default as it did before.

Some observations from testing:
- Power can get stuck at 0% by blowing it up in the vortex from the inside
- The failed demat sound gets cut off at the end
- Force materialising while in the vortex with the doors open results in not being able to walk through the doors until you close and re-open them again
- The explosion effects are visible for every player across the entire map, especially the sparks as they rain down from the sky - maybe set a much shorter TTL on these if we can?
image

lua/entities/gmod_tardis/modules/sh_handbrake.lua
```diff
@@ -1,23 +1,59 @@
...
 	end,
 	serveronly=true,
 	screen_button = {
-		virt_console = false, -- change to true to add
+		virt_console = true,
 		mmenu = false,
 		toggle = true,
 		frame_type = {0, 2},
-		text = "Time-Rotor Handbrake",
+		text = "Time Rotor Handbrake",
```
MattJeanes, line 16: This needs an icon

lua/entities/gmod_tardis_interior/modules/sh_cloisterbells.lua
```diff
@@ -21,6 +21,12 @@ if CLIENT then
 		return self:GetData("cloisters")
 	end)
 
+	ENT:AddHook("ShouldTurnOffCloisters","power",function(self)
+		if not self.exterior:GetData("power-state") then
```
MattJeanes, line 25: This should use GetPower()

MattJeanes (changes requested):

lua/entities/gmod_tardis/modules/sh_lock.lua
```diff
@@ -20,7 +20,7 @@ TARDIS:AddControl({
 		text = "Door Lock",
 		pressed_state_from_interior = false,
 		pressed_state_data = "locked",
-		order = 6,
+		order = 10,
```
MattJeanes, line 23: This control appears to have the same order as the door open - not sure if intentional

lua/tardis/themes/visgui/default-transparent.lua
```diff
@@ -29,6 +29,10 @@ local theme = {
 		scanner = "screen_scanner.png",
 		settings = "screen_settings.png",
 		["TARDIS Status"] = "screen_status.png",
+
```
MattJeanes, line 32: Probably shouldn't be a new line here breaking the two sections up, same below as well

lua/entities/gmod_tardis/modules/sh_lock.lua
parar020100: It is intentional. The order of those doesn't matter, they are technically in the same category. What matters is that they're next to each other and after some certain ones. The sorting supports equal order numbers, I've checked that.

lua/tardis/themes/visgui/default-transparent.lua
parar020100: Fixed

parar020100:
> The explosion effects are visible for every player across the entire map
They are only seen by TARDIS occupants, that is checked. And disabling it in third person would be complicated, since we'd have to check for third person every time we do those effects, it would slow it down, and, imho, not really worth the effort.

> Power can get stuck at 0% by blowing it up in the vortex from the inside
Fixed, now blowing up in the vortex results in falling out of it

> Force materialising while in the vortex with the doors open results in not being able to walk through the doors until you close and re-open them again
Fixed, interrupting the teleport now closes the doors automatically

> The failed demat sound gets cut off at the end
- Couldn't reproduce, it doesn't for me.
-

MattJeanes:
It is specifically the demat_fail_ext.wav that gets cut off, the interior variant is fine.

Blowing up in the vortex already resulted in falling out of it but the health then got stuck at 0% and it was not possible to kill the TARDIS completely afterwards by shooting it more it would just stay in the warning state - is this what you have fixed?

Are you sure it is only visible to occupants? That effect would be visible globally as it's created serverside as part of the explosion - I would recommend using util.Effect / explosion sounds clientside instead and perhaps taking damage directly rather than creating an explosion on the interior

MattJeanes (approved):
I'm happy with these changes, definitely is going to need some testing as these are some core changes to the teleport systems but it's looking pretty good from having a play around. As mentioned I'll await any feedback from you before merging this in but I'm happy to merge if you are.

PR #472, opened 2021-04-10
Exterior cloaking

francodml:
Adds cloaking or "phasing" to the TARDIS.

Works exactly like Legacy's cloaking, with the plus that it can be re-bound to any key, or activated from the virtual console.

## Usage
The TARDIS can be cloaked at any point while there is power and it isn't dead. This can be achieved with the Virtual Console button, a physical console button, or by pressing the keybind (default L) while in third-person flight.

Once cloaked, the TARDIS will be invisible to anyone and everyone, but will remain solid. Opening the door will reveal the interior and the doors as usual.

## Developer stuff
To implement the control, the ID associated with the cloaking control is `"cloak"`. You can specify this in `PART.Control` to use the default handling or implement your own by calling `exterior:ToggleCloak()` or `exterior:SetCloak(bool on)`.

You're allowed to customize your exterior's phase material.
To achieve this, you must create a special material, and a path to said material in the `T.Exterior.PhaseMaterial` key in your interior file. This material can be based off of `materials\models\drmatt\tardis\exterior\phase.vmt`, the only edit you'll need is to point it to your exterior's specific normal map vtf, or any other normal map you wish to use for a phasing effect.

Generally, you may create any material you desire, I suggest basing off of ours because of the effect it creates, but if you've got something else in mind for your particular exterior, feel free to experiment.

The material doesn't need to have the same path as the original phase,vmt, it is instead defined in your interior file as specified above.

## Known issues
There's a strange bug wherein the exterior's material will flicker between the original and the phasing one at the edge of the clipping plane that completely clears off the exterior model. I'm guessing this is due to overlap on the model draws and one single clipping plane being used to cut off both model draws, but I'm not at all sure, so take that with a grain of salt.

Besides this, the clip plane position will flicker slightly while moving up or down too fast. This has been compensated by moving the upper bound of the clip planes past the actual bounding box, but this issue will still occur while mid-phase. This could be due to slow Think-ing or an overhead on the `Set/GetData` functions causing the position data used to calculate clip planes to be slightly behind the actual position of the box, and therefore causing a discrepancy between where the plane should be, and where it's calculated to be.

It has also been pointed out that the way the portal looks when fully cloaked isn't the best, as it has no doorframe, and I do think the cloak effect overall has room for improvement, however I feel as if this could be expanded upon in a future release, once we decide the proper way to go with.


MattJeanes:
It needs an icon for the screen button but otherwise all good, nice work!

MattJeanes (approved):

PR #482, opened 2021-11-08
Visualise TARDIS materialise around player

### LINKED ISSUE #148: Visualising TARDIS teleporting onto player
There should be some kind of effect for when the TARDIS teleports onto you, and you are teleported into the interior. You could use an effect already built into Gmod and that would be cool enough. Coding-wise I imagine this would probably be somewhat simple as all it would have to do is detect that a player was teleported onto and then play the effect. A cool teleportation effect used in another addon is the Buddy Finder. When you teleport there's a pretty cool effect.

parar020100:
Basically, this closes #148

How it looks:
https://youtu.be/j0sFgYzal9w


MattJeanes (approved):
Looks good and works great from my quick testing, nice one!

PR #490, opened 2021-11-10
Refactoring: split sh_teleport.lua into several files

parar020100:
No functional changes, just refactoring


MattJeanes (approved):
Fixed up some weird formatting in sh_tp_problematic.lua not sure if that was in the original code but fixed anyway now, otherwise all looking good!

PR #491, opened 2021-11-11
Sonic dispenser 

### LINKED ISSUE #147: Sonic Screwdriver Dispenser
Like in the S9 finale and the S5 opener, the TARDIS had an opening to deposit Sonic Screwdrivers.

parar020100:
Closes #147

If sonic screwdriver addon is installed, there is a sonic in the console
If it isn't or the model is missing, there's just a button which produces a warning
The control itself can be used in other TARDISes, and will not break the thing if the sonic isn't there
изображение

Made with the allowed partial use of ryanm2711's code in Brundoob's extension

P.S. It currently works in a very simple way. 
If someone wishes to add any effects/animations to it, I'll be more than happy, but I probably won't do that


MattJeanes (approved):
Looks good! Made some tiny tweaks to language but nice stuff

PR #504, opened 2021-11-15
Continue demat if door closed after failed demat

parar020100:
Just a small addition to the update that's been requested several times


MattJeanes (approved):
Looks good and works!

PR #506, opened 2021-11-15
Control messages and fixes

parar020100:
- Added messages for cloak control
- Made many binds use `TARDIS:Control()`
- Fixed float control not being toggled from interior when flightmode is on
- Sonic Screwdriver in default interior is hidden when held in hand


MattJeanes (approved):
Fixed an issue I found with making the sonic invisible but otherwise all good

Issue #500, opened 2021-11-14
Cloaking issue in the vortex


parar020100:
https://cdn.discordapp.com/attachments/812594727018299453/909345609302020156/8mb.video-S2e-0mv2WEoZ.mp4

I think that cloaking shouldn't even be a thing in the vortex.
@MattJeanes @francodml what do you think?

Discovered by Cheaple

MattJeanes:
As discussed in Discord this is an known issue that I tried to fix but there are so many rendering hacks around this code I didn't want to break anything else so just decided it was good enough. The reason we allow it at all is so you can take off and then cloak in the vortex and land cloaked.

Cloaking in the vortex itself is not required so perhaps as a workaround we could make it so that cloaking itself does not activate in the vortex but the control still works so you can do this. Maybe send a message in the user saying cloaking disabled/enabled so they know what it's doing as there is no visual confirmation.

AmyJeanes:
Cloaking appears to work fine now in the vortex. We did do a bunch of rendering fixes in general a while back so sounds like that got it.

<img width="677" height="844" alt="Image" src="https://github.com/user-attachments/assets/f3550c3e-3b39-404d-9c4a-1ab785a6ef67" />

PR #517, opened 2021-11-16
Fixed TARDIS shadow rendering in when cloaked

ryanm2711:


MattJeanes (approved):
This needs to be paired with the fix #500 or we'll get more shadow bugs but otherwise happy with this change

PR #519, opened 2021-11-17
Control and lights last moment fixes

### LINKED ISSUE #454: Some controls can be toggled if power is off
That includes at least hads, isomorphic, but maybe others

It's best if we check all of controls for that. If power is off, as few controls as possible should work (only lock, self-repair + redecoration and materialising if in vortex)

parar020100:
This closes #454

Multiple control-related fixes

- power_independent setting for each control which defines whether it can be
  used when the power is off
- added singleplayer warning for isomorphic
- forbade the use of control sequences when power is off
- replaced some more uses of internal functions with TARDIS:Control()
- fixes for screens (ply -> LocalPlayer(); added control activator where possible)
- added part option PART.PowerOffSound (whether PART.Sound is played with no power)
- added part option PART.PowerOffUse (whether PART can be used at all with no power)

\+ fixed the rendering of the turned off lights


MattJeanes (approved):
Looks good to me!

Issue #214, opened 2016-11-04
Uncloak sound


Noname12345GER:
What if you would be able to play a reversed cloak sound on uncloak?

MattJeanes:
If someone finds me a good enough sound for it sure

kennyrkun:
[I made this. It's probably terrible.](http://www.mehsworld.com/p/kennyrkun/files/phase_disable.wav)

doctorwho1200:
Well, i have one
Took it from the episode and reversed
But i've no idea how to put it in here :P

kennyrkun:
@doctorwho1200 put it in a zip file and upload that, or put it on YouTube and link it.

doctorwho1200:
bah, too complicated. this should work
https://www.dropbox.com/home?preview=uncloak.wav

Noname12345GER:
The Doctor included all sounds in his Extensions for the case of being implemented.

MattJeanes:
@doctorwho1200 yeah thats a cool sound when I get round to putting cloak in the rewrite i'll use that

Issue #267, opened 2017-05-23
Door


ProfessorNAW:
The camera on the door doesn't work. The right door from the inside isn't attached to the TARDIS. The info box on the door isn't just on the outside and is sticking out.

ProfessorNAW:
You also can't shoot rockets or bullets through the door.

ProfessorNAW:
The TARDIS interface will pop up when you are done flying.

ProfessorNAW:
You can walk straight through the doors while the doors are closed.

MattJeanes:
Please raise these all as separate issues, it's impossible to track them as one.

PR #549, opened 2021-12-01
Cloak and interior lock sounds

parar020100:


parar020100:
Closes #214 #267

MattJeanes (approved):
I boosted the sound up for cloak disengage a bit but otherwise good. The interior lock sound seems to have some sonic in the background but not sure we can do about that. Maybe worth thinking about but good enough.

Issue #448, opened 2021-03-22
Allowing classic style doors to be possible


FuzzyLeo:
In the show the only 'canonical' appearance of the view through the classic doors has shown us that the police box doors are completely absent from the interior, now no matter how much anyone may dislike this, its how it is in both the classic and reboot of the show and therefore should at least be a possibility for those who _do_ want it.

Pictures from the show:
![image](https://user-images.githubusercontent.com/23579757/112070011-cac79500-8b64-11eb-9241-cb073ac92dbf.png)
![image](https://user-images.githubusercontent.com/23579757/112070026-d1560c80-8b64-11eb-887b-9ed83fd426e2.png)
![image](https://user-images.githubusercontent.com/23579757/112070046-d915b100-8b64-11eb-87d3-3900147a2cab.png)
![image](https://user-images.githubusercontent.com/23579757/112070182-19752f00-8b65-11eb-87db-1b6e709e0466.png)

The most plausable way it could be done from the interior is when you're inside the interior it would make the exterior invisible, producing an effect like this (assuming the portal has been adusted in the rewrites code):
![image](https://user-images.githubusercontent.com/23579757/112070491-aa4c0a80-8b65-11eb-8aa1-6759b6dff478.png)

and from the exterior it would be done by allowing the exterior doors to render on top of the portal (which also solves the issue of the portal seam being visible), which would look something like this:
![image](https://user-images.githubusercontent.com/23579757/112071321-6823c880-8b67-11eb-9015-ca3982c4861d.png)
![image](https://user-images.githubusercontent.com/23579757/112071338-72de5d80-8b67-11eb-9342-cc976ae5e563.png)

These two methods combined would make it possible for classic style doors to be created with the added benifit of making the portal a huge amount less noticable.

parar020100:
Are different-sized portals already supported?

francodml:
Should be. We just need to tweak some rendering code.

MattJeanes:
@parar020100 yes they are, you just set different sizes in exterior vs interior in the metadata - the tricky bit is rendering a door over the portal

MattJeanes:
@francodml how you feeling about this one? I'm happy to close if you're not working on this one

francodml:
> @francodml how you feeling about this one? I'm happy to close if you're not working on this one

I'm not currently working on this due to time constraints on my end. I might start looking into it once I'm free in a few weeks.

parar020100:
[...]

PR #550, opened 2021-12-01
Classic style doors (#448)

parar020100:


lua/entities/gmod_tardis_interior/modules/sh_portals.lua
```diff
@@ -44,7 +44,7 @@ else
 		end
 		if dont then
 			return false, black
-		elseif (not (self.DoorOpen and self:DoorOpen(true))) and portal==self.portals.interior then
+		elseif (not (self.DoorOpen and self:DoorOpen(false))) and portal==self.portals.interior then
```
MattJeanes, line 47: What was the reason for this change?

lua/entities/gmod_tardis/modules/cl_classic_doors.lua
```diff
@@ -0,0 +1,35 @@
+local function SetVisibleIfExists(part, visible)
```
MattJeanes, line 1: This should ideally be replaced with a ShouldDrawPart hook, I thought one existed but it doesn't, but it should be a pretty trivial change to make in the parts system [around here](https://github.com/MattJeanes/TARDIS/blob/4f2d7937f153faa20537c7777f4e2fa176ba4935/lua/tardis/sh_parts.lua#L12-L18). Spamming `SetRenderMode` / `SetColor` hooks can cause frame drops and here it is not necessary as we have a better and cleaner solution.

MattJeanes (changes requested):
Overall looking pretty good, but I have a better approach for actually making the parts invisible which I have detailed in a comment which I think we should definitely go for. Happy to help out if you need me to on that.

lua/entities/gmod_tardis_interior/modules/sh_portals.lua
parar020100: Interior door closes slower
This prevents the interior portal from disappearing before it's closed

MattJeanes: Okay, hopefully this has no negative side effects but I suppose testing will reveal that, just got to be careful messing with core rendering code

lua/entities/gmod_tardis/modules/cl_classic_doors.lua
parar020100: Should be fixed

lua/entities/gmod_tardis/modules/cl_classic_doors.lua
```diff
@@ -0,0 +1,29 @@
...
+
+end)
+
+ENT:AddHook("ShouldDrawPart", "classic_doors_support", function(self, part)
+	if self.metadata.EnableClassicDoors ~= true then return end
+	if part == nil then return end
+
+	local ply = LocalPlayer()
+
+	local inside = (ply:GetTardisData("exterior") == self
```
MattJeanes, line 20: We should not be checking player state as this will quickly break down if the player is using a camera view or something else. We should be using wp.drawing + wp.drawingent to determine if we're currently drawing the appropriate portal and hide it in that context instead. Same with the other instances where we are checking these variables.

lua/tardis/sh_parts.lua
```diff
@@ -8,15 +8,24 @@ function TARDIS.DrawOverride(self,override)
...
-		if (self.InteriorPart and IsValid(int) and ((int:CallHook("ShouldDraw")~=false)
-		or (ext:DoorOpen()
-			and (self.ClientDrawOverride and LocalPlayer():GetPos():Distance(ext:GetPos())<TARDIS:GetSetting("portals-closedist"))
-			or (self.DrawThroughPortal and (int.scannerrender or (IsValid(wp.drawingent) and wp.drawingent:GetParent()==int)))
-		))) or (self.ExteriorPart
-			and (ext:CallHook("ShouldDraw")~=false)
-			or self.ShouldDrawOverride
-		) then
+
+		local dist_to_ext = LocalPlayer():GetPos():Distance(ext:GetPos())
```
MattJeanes, line 14: The reason this was done inline was to optimise calls to this so it is only called when required, can we put this back inline?

MattJeanes (changes requested):

lua/tardis/sh_parts.lua
parar020100: Done

lua/entities/gmod_tardis_interior/modules/cl_classic_doors.lua
```diff
@@ -0,0 +1,20 @@
...
+	if self.metadata.EnableClassicDoors ~= true then return end
+	if part == nil then return end
+
+	local ply = LocalPlayer()
+
+	local inside = (ply:GetTardisData("interior") == self
+		and not ply:GetTardisData("thirdperson")
+		and not ply:GetTardisData("destination"))
+
+	--TARDIS:Debug("A")
```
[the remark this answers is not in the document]
parar020100: Done

lua/tardis/sh_parts.lua
parar020100: Done

lua/entities/gmod_tardis/modules/cl_classic_doors.lua
parar020100: Should be resolved

lua/entities/gmod_tardis/modules/cl_classic_doors.lua
```diff
@@ -0,0 +1,29 @@
+ENT:AddHook("ShouldDraw", "classic_doors_support", function(self)
+	if self.metadata.EnableClassicDoors ~= true then return end
```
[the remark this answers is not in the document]
parar020100: Should be resolved

MattJeanes (approved):

parar020100:
Closes #448

Issue #539, opened 2021-11-30
Change long flight control name


parar020100:
"Vortex Flight Toggler" just sounds wrong
It should at least just be "Vortex Flight"

Issue #537, opened 2021-11-30
TARDIS MSG TYPE 2 ORDER


parar020100:
"ERROR: [TARDIS] ..." seems wrong
I think it should be "[TARDIS] ERROR: ..."

Issue #527, opened 2021-11-23
Light, shield and stabilizer control placeholders


parar020100:
Related to #373

Shield might be required for some other global changes, like #371, #375, #372
Lights: #492
Stabilizers: #536

We should add placeholders that extensions would assign the part to, so that it would work when added
These are very likely to be added in the future

Issue #485, opened 2021-11-09
Reduce flight speed with opened doors


parar020100:
I think this would improve the gameplay
As we know from "The Runaway Bride", the TARDIS isn't very good at flying with open doors

Brundoob:
I don't think it's a case of it being hard to fly with doors open, as much as it was trying to fly it from the doorway instead of the control panel. The 12th Doctor and Clara were mapping out london and he was flying with no trouble with open doors

MattJeanes:
Maybe if the doors are open we should automatically disable spinning at least, that might be a good tweak to make

parar020100:
> I don't think it's a case of it being hard to fly with doors open, as much as it was trying to fly it from the doorway instead of the control panel. The 12th Doctor and Clara were mapping out london and he was flying with no trouble with open doors

12 and Clara were flying very very slowly

parar020100:
As for flying from the doorway... it relates to #374, I'll think of a way to do both

PR #553, opened 2021-12-05
Spin and flight fixes, some control placeholders

parar020100:
Gameplay changes: (#485 and others)

- Boosted (with shift) flight is a bit faster than it used to be if TARDIS spinning is enabled (spin dir ~= 0)
- Boosted (with shift) flight is a bit slower than it used to be if TARDIS spinning is disabled (spin dir = 0)
- Flight boost control (shift) slows down the TARDIS instead of boosting if its doors are open 
  - can be configured, **ENABLED** by default
  - a warning message telling to close the doors
- Spinning now stops when the TARDIS doors are opened in flight
  - can be configured, **DISABLED** by default
  - after doors are closed, returns back to normal

Fixes and internal additions:
- #537 (error message order) : fixed
- #539 (control name) : fixed
- #441 (spin direction control) and #442 (separate controls for toggle spin and spin direction): done 
  - Three controls: **spin_cycle, spin_toggle, spin_switch**
  - NOTE: when this is released, we should add those new controls to wiki
- #527 (control placeholders for extension devs): added
  - NOTE: when this is released, we should add those new controls to wiki


MattJeanes:
I am happy for the doors open stopping spinning to be enabled by default. However for the boost slow down with doors open I do not really understand why this was added? It seems to serve no purpose and I can see this causing a lot of frustration and annoyance so would at least like this set to off by default.

Also you mention in #442 about having the atom accel toggle it and directional pointer modify direction on the default interior which doesn't appear to be the case here, although I am not sure if we should be changing this feature at least without some kind of hint if they try pressing it multiple times in a short period about where the control has gone maybe.

MattJeanes:
Code itself looks good and appears to work as intended ingame just a few concerns around the logic as above.

MattJeanes:
Actually have just noticed that now we seem to spin faster when flying normally and slower when boosting, is this intended behaviour with these changes? It feels very strange in game - I don't think we should by modifying the default flight mechanics in this way

MattJeanes (approved):

PR #595, opened 2022-01-19
Migrate music while redecorating

CrethFox:
- The music doesn't stop when you redecorate
- It can be controlled from the new interior

### DIFF
--- a/lua/entities/gmod_tardis/modules/cl_music.lua
+++ b/lua/entities/gmod_tardis/modules/cl_music.lua
@@ -98,3 +98,7 @@ ENT:AddHook("OnRemove", "music", function(self)
 		self:StopMusic()
 	end
 end)
+
+ENT:AddHook("MigrateData", "music", function(self, parent)
+	self.music = parent.music
+end)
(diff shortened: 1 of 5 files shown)


AmyJeanes (commented):
Okay so I like what you're doing here but I'd like to see this be done a bit more in a generic way so other things can make use of this. I think if you use the [message module](https://github.com/MattJeanes/TARDIS/blob/main/lua/entities/gmod_tardis/modules/libraries/sh_message.lua) rather than directly sending network messages and make it generic instead of exclusively music then you can use `self:CallHook("MigrateData", parent)` or something both on the server and client.

Then other in your music module you can add a hook for MigrateData and move the music across like that e.g.

```lua
ENT:AddHook("MigrateData", "music", function(self, parent)
    self.music = parent.music
end)
```

If you need some help with this feel free to reach out, appreciate this might make little sense! 

Search "flash-light" in this file for an example of the message library: https://github.com/MattJeanes/TARDIS/blob/cc904cc891f6b6b9bd36a50e032e0879f44d14d9/lua/entities/gmod_tardis/modules/sh_light.lua

AmyJeanes (approved):
I've made a fix to the `MigrateData` hook cause when the client receives the message it may not actually know about the new exterior yet due to network timing, so now we wait for data load (new hook `DataLoaded` on the client) and check the synced parent data there. Should be more reliable. God knows why the network messages aren't sent in some sort of sensible and predictable order 🤷🏻 

I'm happy with it now but would like @parar020100 for you to go over these changes yourself and test them if possible to see if you're happy also. If so, I'm happy for you to merge.

PR #605, opened 2022-01-20
Open third person with ALT (WALK) key

### LINKED ISSUE #596: Change "third person" keys: "Alt+E" instead of "E" (on the console)
This is a solution to #184 which seems to be simple enough, so I'll close that one

* alt+E on the console => third person
* E on the console => a notification message telling to use alt to open third person 
* an option to disable this; from @MattJeanes:
  > i think if it comes with an option to revert to normal behaviour that should be fine
i can see people moaning about it

parar020100:
Closes #596


lua/entities/gmod_tardis/modules/sh_thirdperson.lua
```diff
@@ -13,6 +13,17 @@ TARDIS:AddKeyBind("tp-toggledoor",{
 	exterior=true
 })
 
+TARDIS:AddSetting({
+	id="thirdperson_careful_enabled",
+	name="Use walk key to open third person",
+	desc="Should the WALK ('ALT' by default) key be pressed to enter third person by clicking the console?",
```
AmyJeanes: Recommend minor language improvement here, not sure how this will look ingame though

```suggestion
	desc="Should the WALK ('ALT' by default) key be pressed to enter third person when pressing USE ('E' by default) key on the console?",
```

AmyJeanes (approved):

PR #607, opened 2022-01-29
Several small fixes

### LINKED ISSUE #518: Reconsider engine release behaviour
From discussion with Blazi/Creth:

> Emergancy landing with the handbreak during vortex flight caused it to materialise instantly
> the remat sound is way too long if at all required for an instant remat
> if the sounds are accurate you can step out the door before its even done rematting

> This is intentional. This is taken from season 2, "Rise of the Cybermen"
> It is about TARDIS falling out of the vortex, not materialising like it normally would

> well congradulations it broke my immersion entirely 
> the remat sound is way too long if at all required for an instant remat
> if the sounds are accurate you can step out the door before its even done rematting

parar020100:
Important:
  - Fixed classic door collisions
  
Rotor:
  - A slight codestyle improvement (tabs)
  - Removed logically unrequired conditions
  
Tips: 
  - Reduced the size of the tip frame (as requested by many people with lower screen resolution)

Gameplay:
  - Fixed #518 (interrupted teleport sounds)
  - Made flight speed and spin dependent on `doorstatereal` rather than `doorstate`
    - the TARDISes used to wait for classic doors to close for normal flight, it was horrible
  
Default interior:
  - Added some part sounds
  - Replaced some buttons with levers (as noticed by some players, the old version could be seen as inappropriate)


lua/entities/gmod_tardis/modules/cl_classic_doors.lua
```diff
@@ -14,4 +14,23 @@ ENT:AddHook("ShouldDrawPart", "classic_doors_exterior_door", function(self, part
...
+	if not IsValid(intdoor) then return end
+
+	local door_sounds = self.metadata.Interior.Sounds.Door
+	if not door_sounds then return end
+
+	local door_sound = self:GetData("doorstatereal") and door_sounds.open or door_sounds.close
+	if not door_sound then return end
+
+	if intdoor.IntDoorPos ~= nil and intdoor.IntDoorPos ~= 0 and intdoor.IntDoorPos ~= 1 then
+		sound.Play(door_sound, self.interior:LocalToWorld(Vector(0,0,0)))
```
AmyJeanes: Any reason why we have LocalToWorld(Vector(0,0,0)) vs just GetPos()? In this context they are identical, also should the door sound be coming from the centre of the interior?

lua/tardis/interiors/default.lua
```diff
@@ -71,7 +71,7 @@ T.Interior = {
 		default_screen             = {pos = Vector(42, 0.75, 27.1),                   ang = Angle(0, -5, 0),                  },
 		default_screenbutton       = {pos = Vector(44.5, 9.75, 38.2),                 ang = Angle(215, 85, 90),               },
 		default_wibblylever        = {pos = Vector(-48, 18, 5.4),                     ang = Angle(335, 347, 6),               },
-		default_enginerelease      = {pos = Vector(-61.229, 21.622, -4.514),          ang = Angle(90, 90.911, -90),           },
+		default_enginerelease      = {pos = Vector(-61.229, 21.622, -4.514),          ang = Angle(180, 90.911, -90),           },
```
AmyJeanes: nitpick: new number pushes the alignment out of the closing bracket

lua/tardis/interiors/default.lua
```diff
@@ -93,7 +93,7 @@ T.Interior = {
 		default_sonicdispenser     = {pos = Vector(-26.048, 42.31, 1.45),             ang = Angle(9.445, 46.482, -21.861),    },
 		default_helmic             = {pos = Vector(-26, -41, 4),                      ang = Angle(0, 330, 24.5),              },
 		default_sonic_inserted     = SonicModelExists() and
-	                                 {pos = Vector(-29.748, 40.71, -20.15),           ang = Angle(-100, -90, 90),             },
+									 {pos = Vector(-29.748, 40.71, -20.15),           ang = Angle(-100, -90, 90),             },
```
AmyJeanes: nitpick: new alignment pushes the alignment out of the closing bracket

parar020100: In this case, it does not; I just replace spaces with TABs, github is displaying it the wrong way

lua/entities/gmod_tardis/modules/sh_doors.lua
```diff
@@ -99,6 +99,15 @@ if SERVER then
 		end
 	end)
 
+	ENT:AddHook("ToggleDoorReal", "classic_doors_collision", function(self,open)
+		if self.metadata.EnableClassicDoors ~= true then return end
+
+		local int_classic_door=TARDIS:GetPart(self.interior,"intdoor")
+		if IsValid(int_classic_door) then
+			int_classic_door:SetCollide(not open, true)
```
AmyJeanes, line 107: What's the purpose of this for? In the hell bent one I've been testing this with in-game, you can walk through the interior doors now (not sure if you could before to be fair) when the doors are open. I'm just it's fine just more of a curiosity

lua/entities/gmod_tardis/modules/cl_classic_doors.lua
parar020100: Left it there accidentally
I've now changed it to use fallback coordinates instead

lua/tardis/interiors/default.lua
parar020100: Fixed

lua/tardis/interiors/default.lua
AmyJeanes: GitHub shows the discrepancies between space and tabs when they are both used in the same file. Only reason it looks fine in your editor is the tab width is different.

You can see here these is a mixture of tabs and spaces even on the same line for most of these. I don't mind which we use but it should be consistent within the codebase ideally or at least the file

image

lua/entities/gmod_tardis/modules/sh_doors.lua
parar020100: It's the way portals work
If you can go through normally, you get to the portal exit
If after getting out of the portal you get stuck, you get teleported to Fallback pos, which is okay but not as smooth/precise

You can test it by setting the fallback to a different side of the interior
alt+click would take you to fallback
getting stuck while teleporting would take you to fallback, as it used to happen 
now, the interior doors don't have the collision => you don't get teleported to fallback, you just come out

parar020100: Basically, classic doors used to always teleport you to fallback

AmyJeanes: Ah okay, gotcha

AmyJeanes (approved):

PR #609, opened 2022-01-30
Time distortion generator 

### LINKED ISSUE #122: Stop teleporting into certain locations
This would be perhaps a tool usable by people to maybe whitelist or blacklist certain people from entering an area. By nature the tool would likely be admin-only.

parar020100:
Closes #122

## Time distortion generator ("TDG" in further references)

I've tried to make it cause as many complications to players as possible (to make the game more interesting)

- If a TDG is activated outside the TARDIS:
  - it causes failed demat
  - you can still force demat with engine release
  - force demat causes the TDGs near the TARDIS explode
    - TDGs will explode in a small radius (300), but the working radius of TDG is 1000
      (so you have to find it if it's hidden)

- If a TDG is activated near the selected destination:
  - The TARDIS can not land there
  - Engine release and other things won't help - landing **IS** impossible
  - If you cannot land anywhere (for example, on a small map), you can use the handbrake to fall out
    - This won't make you get stuck
  - If vortex flight is on, you will stay in the vortex with failed mat sound
  - If vortex flight is off, you will rematerialise back where you dematerialised from

- If a TDG is activated inside the TARDIS:
  - You can demat, but each demat will be like force-demat
    - sparks, shaking and damage as usual :)
  - Each teleport will have its coordinates randomized 
  - Good luck finding it in huge corridor systems like 2010

- If TDGs are activated inside and outside the TARDIS at the same time:
  - You cannot take off
  - Even with engine release
  - Don't even try it
  
**LET THE WARS BEGIN!**

## Other fixes:
- Removed duplicating code (which still exists in sh_tp_vortex.lua)
- Much refactoring of the teleport feature
- Replaced some occurences of timer.Simple() with ent:Timer()
- Removed exterior-specific failed demat sound
  - It had been pointed out by many people that it has toyota interior sounds in it
  - There's only one failed demat sound by default (extensions can still customize)


AmyJeanes (approved):
Another fantastic update, I'm sure people are gonna have loads of fun with this! You probably know the teleport code better than me at this point lol

PR #610, opened 2022-01-31
Interior versions and other changes

parar020100:
- interior versions
  - split "classic doors" and "double doors" versions
    - added a setting for the player to choose the preffered one
  - allowed custom versions
  - behaviour discussed in Discord already
- interior preferences
  - you can select favorite interiors, they will be on top of the list
  - for each interior, you can set whether it can be selected randomly during redecoration
- interior templates
- support for custom interior hooks
- support for overriding tip texts for interiors
- allowed `TARDIS:Debug()` during game loading
- refactored category and name overrides
- added texture sets and `ENT:ApplyTextureSet(set_id)` function for use in the interiors

изображение


AmyJeanes (approved):
Very nice work! I've tried it with the hell bent extension dev branch and all working good, combined with the design you showed me before this looks great!

PR #612, opened 2022-02-04
TT Capsule version of default interior and other fixes

parar020100:
- Removed "default interior" setting and related GUI elements

- Made templates merge after all interiors are added

- Added some CustomHook{} safeguards

- TT Capsule version of default interior
  - not randomized
  - a TARDIS template with the exterior (for extensions to use)
  
- Added the support of winter (snowy) TARDIS skins
  - they should not be selected randomly unless it's enabled in the settings
  
- **Fix**: part-assigned tips only show for existing parts
  - required for interior versions with different levers (like 1965 extension)
  
- **Cleanup**: moved default interior support into metadata.Interior.CustomHooks{}
  - it now works for versions of it as well
  
- **Troll**: when a TDG is inside the TARDIS, some tip texts are randomized
  - 5 people who already saw it asked you to allow me keep this
  - to prevent confusion, occupants and people who enter the TARDIS get notified about it


AmyJeanes (approved):
Looks good to me!

PR #614, opened 2022-02-04
Trigger HADS when lifted by physgun

### LINKED ISSUE #559: Activate HADS is lifted with a physlock
This is a sort of attacking (?)
Unsure

DoctorErt:
Closes #559

### DIFF
--- a/lua/entities/gmod_tardis/modules/sh_hads.lua
+++ b/lua/entities/gmod_tardis/modules/sh_hads.lua
@@ -10,7 +10,7 @@ if SERVER then
         return self:SetHADS(on)
     end
 
-    ENT:AddHook("OnTakeDamage", "hads", function(self)
+    function ENT:TryTriggerHADS()
         if self:CallHook("CanTriggerHads") == false then return end
         if (self:GetData("hads",false) == true and self:GetData("hads-triggered",false)==false) and (not self:GetData("teleport",false)) then
             if self:GetData("doorstatereal") then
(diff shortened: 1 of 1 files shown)


lua/entities/gmod_tardis/modules/sh_hads.lua
```diff
@@ -10,7 +10,7 @@ if SERVER then
         return self:SetHADS(on)
     end
 
-    ENT:AddHook("OnTakeDamage", "hads", function(self)
+    function ENT:TryTriggerHADS()
```
AmyJeanes: This can be just TriggerHADS like the other functions that work similarly

lua/entities/gmod_tardis/modules/sh_hads.lua
```diff
@@ -10,7 +10,7 @@ if SERVER then
...
         if self:CallHook("CanTriggerHads") == false then return end
```
AmyJeanes: Can we change up the logic in this function so that it returns true or false whether or not HADS was activated? This will help below

lua/entities/gmod_tardis/modules/sh_hads.lua
```diff
@@ -34,6 +34,19 @@ if SERVER then
...
         end
+    end 
+
+    ENT:AddHook("OnTakeDamage", "hads", function(self)
+        self:TryTriggerHADS()  
+    end)
+
+    hook.Add("OnPhysgunPickup", "tardis-hads", function(ply,ent)
+        if ent:GetClass()=="gmod_tardis" then
+            ent:TryTriggerHADS()
```
AmyJeanes: As above when TriggerHADS returns true/false we can use that in the if check below rather than checking hads-triggered here

AmyJeanes (changes requested):
Hey there, thank you for your contribution! I see this is your first ever GitHub contribution so congratulations! 😄 

I've added a couple little comments to tweak the behaviour slightly but this is looking pretty good!

DoctorErt:
Everything should be fixed now

AmyJeanes (approved):
Many thanks for your contribution!

PR #615, opened 2022-02-04
Random vortex position

### LINKED ISSUE #533: Fall out at a random place after undo-ing TARDIS in vortex
Currently, when you just get back to the last position of the TARDIS, it's cheaty and game-breaking (allows to return back quickly)

Falling out at a random place would fix that

DoctorErt:
Closes #533


parar020100:
This should also fix the bugs where parts of the world are seen in the vortex

изображение

AmyJeanes (changes requested):
Hey thanks for the contribution, so the code looks good but in-game the TARDIS suddenly jumps up into the sky before the vortex has time to fade in so it doesn't provide a smooth experience anymore as it did before. The vortex fades in over about 2 seconds clientside so to allow time for networking and lag I'd recommend adding a 3 second delay to this teleport.

parar020100:
Migrated to #623

PR #618, opened 2022-02-08
Hads auto remat and forbid respawn for non-owners

### LINKED ISSUE #369: Random location in HADS
Currently, the TARDIS just drifts away into vortex if attacked. Unless it is linked to a sonic, which a player may forget to do, the TARDIS is lost.

Suggestion: the TARDIS should travel to a random location instead (like in "Cold War")

### LINKED ISSUE #580: Guests shouldn't respawn inside the TARDIS
Only the owner should be able to respawn inside.
If #532 is done, the companions also could.

That would prevent unwanted guests from ruining the TARDIS experience (you'd just kill them to get them out)

DoctorErt:
Closes #369
Closes #580


AmyJeanes (approved):
Looks good, nice work!

PR #619, opened 2022-02-08
Custom settings

parar020100:
Added custom interior-specific settings
It also allows to choose which interiors will spawn

изображение

Preferred version submenu:
изображение


lua/entities/gmod_tardis_part/shared.lua
```diff
@@ -37,4 +37,18 @@ end)
 
 hook.Add("CanDrive", "tardis-part", function(ply,ent)
     if ent.TardisPart then return false end
-end)
\ No newline at end of file
+end)
+
+function ENT:SetVisible(visible)
```
AmyJeanes: Why has this been moved to shared? It can only effectively work serverside (unless you're spamming it every frame clientside which is bad for performance)

lua/tardis/sh_interiors.lua
```diff
@@ -76,47 +96,130 @@ function TARDIS:SaveIntPreferences(preferences)
...
+----------------------------------------------------------------------------------------------------
+-- Interior custom settings
+
+function TARDIS:GetMainVersionId(int_id)
+    return (self.Metadata[int_id] and self.Metadata[int_id].IsVersionOf) or int_id
 end
 
-function TARDIS:IsUnwantedInt(id, ply)
-    return TARDIS:GetIntPreferences(ply).unwanted[id]
+function TARDIS:GetIntCustomSetting(int_id, setting_id, ply, default_val)
```
AmyJeanes: Can we rename these functions to use the full Interior word? I think the use of `Int` here is ambiguous. E.g. `TARDIS:GetInteriorCustomSetting`. Fine to use inside the actual function but the "public" API should be clear

lua/tardis/sh_interiors.lua
```diff
@@ -76,47 +96,130 @@ function TARDIS:SaveIntPreferences(preferences)
...
+    end
+
+    if setting_id == "preferred_spawn" then
+        return TARDIS:SetupPreferredSpawn(int_id, ply)
+    end
+
+    local md_settings = metadata.CustomSettings
+    if md_settings and md_settings[setting_id] and md_settings[setting_id].value ~= nil then
+        local metadata_value = md_settings[setting_id].value
+        TARDIS:SetIntCustomSetting(int_id, setting_id, metadata_value, ply)
```
AmyJeanes: This is very confusing, why is a Get method calling the Set method?

lua/tardis/sh_interiors.lua
```diff
@@ -76,47 +96,130 @@ function TARDIS:SaveIntPreferences(preferences)
...
+end
+
+function TARDIS:ToggleIntCustomSetting(int_id, setting_id, ply)
+    local int_id = TARDIS:GetMainVersionId(int_id)
+
+    local value = TARDIS:GetIntCustomSetting(int_id, setting_id, ply)
+    TARDIS:SetIntCustomSetting(int_id, setting_id, (not value), ply)
+end
+
+function TARDIS:SetupPreferredSpawn(int_id, ply)
```
AmyJeanes: Maybe `TARDIS:SetupPreferredVersion` would make more sense?

lua/tardis/sh_interiors.lua
```diff
@@ -226,33 +337,132 @@ function TARDIS:AddInterior(t)
     end
 end
 
-function TARDIS:MergeTemplates()
-    if not self.Metadata then return end
+local function AddInteriorPartsOffset(template, offset)
```
AmyJeanes: In what scenario would all the tips need to be offset from a certain amount?

lua/tardis/sh_interiors.lua
```diff
@@ -154,9 +251,22 @@ local name_overrides = TARDIS_OVERRIDES.Names or {}
 
 local default_category = TARDIS_OVERRIDES.MainCategory or "Doctor Who - TARDIS"
 
-TARDIS.MetadataTemplates = TARDIS.MetadataTemplates or {}
+function TARDIS:FullReloadInteriors()
```
AmyJeanes, line 7: Could you describe why this is needed? The code should be designed in such a way that auto reload can reload everything automatically without having to do a full reset like this every time. It's fine if it's required, but I want to make sure we're not hiding a problem elsewhere by doing this

lua/tardis/sh_interiors.lua
```diff
@@ -367,6 +560,145 @@ local function AddVersionSubMenu(dmenu, version)
...
+        end
+
+        if versions.main.classic_doors_id then
+            add_version_option(versions.main.classic_doors_id, "  Classic doors version") -- spaces are intentional (sorting)
+            add_version_option(versions.main.double_doors_id, "  Double doors version")
+        end
+
+        for k,v in SortedPairs(versions.other) do
+            if v.classic_doors_id then
+                add_version_option(v.classic_doors_id, "  " .. v.name .. " (classic doors)")
```
AmyJeanes: Why does these other / custom versions have spaces in front of them as well? If everything has spaces doesn't that kind of defeat the point?

lua/tardis/sh_interiors.lua
```diff
@@ -529,8 +864,4 @@ if SERVER then
     end)
 end
 
-TARDIS:LoadFolder("interiors/templates", nil, true)
-TARDIS:LoadFolder("interiors", nil, true)
-TARDIS:LoadFolder("interiors/versions", nil, true)
-
-TARDIS:MergeTemplates()
\ No newline at end of file
+TARDIS:FullReloadInteriors()
```
AmyJeanes, line 61: This file is now very very long, would it make sense to move some of this stuff out to a separate module if possible?

AmyJeanes (changes requested):
This looks like it works in-game as far as I can test, but the interiors I have installed do not use some of the new features here. Could you send me something I can use that tests out these new settings + template rotation / offset stuff so I can give it a go?

I've added some questions / comments above in my review, but overall looking really nice - although I am concerned about the much increased code complexity in this area now.

lua/entities/gmod_tardis_part/shared.lua
parar020100: It created some confusion in extensions
The second most common use-case for this is in the custom hooks, having to specify `if SERVER then` just for that each time is not convenient
I've made it accessible from "shared" code just to prevent errors and confusion

lua/tardis/sh_interiors.lua
parar020100: The lines would get too long
They are already
It literally wouldn't fit in a the simpliest "if" clause 
Hm... I think I'll just make it `GetCustomSetting()`

изображение

lua/tardis/sh_interiors.lua
parar020100: This is the default-value clause
If the setting is not found, it is being recieved from default value of metadata
If so, we save the value of the setting to be that default value
That way, we avoid a bunch of unnesessary checks each time it's spawned

lua/tardis/sh_interiors.lua
parar020100: I warned you about this in DMs
They are not all spaces

Some are spaces, some are `alt+255` symbols (aka "unbreakable gaps")
Some are a combination of a space and this symbol (thus, two symbols in each instance)

We keep the options sorted, but also aligned vertically

lua/tardis/sh_interiors.lua
parar020100: - Interiors are not only dependent on each other, but also on templates
- The templates need to be inherited from t.Base, and also overridable
  - We need to be able to disable the parent's template with `<template_name> = false`
  - Thus, they have to be merged strictly after all `t.Base` tables are merged
  - That's why template merging can not be in AddInterior()
- When you changed the interior, AddInterior() would not call merging the templates
- When you changed the template, without reloading the interior it would completely wreck up the interior

I tried to make an automatic re-merge of this interior in AddTemplate() and AddInterior()
I failed. 

One of the "game-freaks-out" scenarios: 
- Interiors `A` and `B`
- `A` has templates `a1`, `a2` and `a3`; `a3` is an overriding template, `a1` and `a2` are not
- `B.Base=A`
- B inherits `a1`, `a2`, `a3` templates from metadata
- We specify that `a2 = false` in `B` (it should not happen), and add `b2` template

Currently, interiors have chains of 3-4 `T.Base` merges
This whole thing is too complicated to track all the merges and re-make them in the right order

lua/tardis/sh_interiors.lua
parar020100: The main thing templates were made for is being able to use one console in several TARDISes.
(MCCC is the one that will work like that)

We want to be able to move that console, and also rotate it, for the specific interior it's being used for.

Ruth TARDISes are practically the same interior, but the console was rotated differently in them.
Same for 1963 - I want an "Unearthy Child" rotation without having to rewrite the coords of the same thing

Plus, obviously, the pranks: 😈 
изображение

изображение

lua/tardis/sh_interiors.lua
parar020100: Renamed into GetCustomSetting()

lua/tardis/sh_interiors.lua
parar020100: I was afraid to cause confusion there, since a `version` is already a thing 
(as "versions" which can be chosen on the right in the spawnmenu, either a named id or a combination of two IDs)

parar020100: Nevermind, it fits, I renamed it

lua/tardis/sh_interiors.lua
parar020100: Sure
I've added `interiors/modules` subfolder for that and split the code by functionality


изображение

lua/tardis/sh_interiors.lua
parar020100: Actually, I've double checked - it's not required any more
(was left there due to dozens of refactoring changes)
It's removed now

parar020100:
I've replied about testing versions in Discord

lua/tardis/sh_interiors.lua
parar020100: UPD: Interiors and templates which don't require this can now specify `.NoFullReload` to prevent this

Fixed in c4e36bf08e61f15152a2894bfd352227a91c02fb

lua/tardis/sh_interiors.lua
AmyJeanes: Oh okay I see

lua/tardis/sh_interiors.lua
AmyJeanes: Okay, considering this only happens when you've developing interiors anyway and not under normal circumstances it's not a huge deal, just wanted to make sure this wasn't just added as a get out of jail free card, when it could be fixed. Well maybe it still can but this seems like an okay alternative.

lua/tardis/sh_interiors.lua
AmyJeanes: Okay, sounds good!

lua/entities/gmod_tardis_part/shared.lua
AmyJeanes: I disagree with this, I think having it in shared code could cause even more problems and confusion in that SetColor does not persist clientside unless you're spamming it every frame which is also a big performance hit. I'm not sure under what scenarios you'd generally need to specify `if SERVER then` anyway here, as the Use code will generally be called serverside only, and as I mentioned we should not be calling this in a clientside Think hook.

lua/tardis/sh_interiors.lua
AmyJeanes: Works for me

AmyJeanes (approved):
Looking good in-game! Tested with development/dev branch of 1963 + Ruth extensions

PR #620, opened 2022-02-10
New interior sounds on damage

CrethFox:
When Damage occurs to the TARDIS the screen shakes a bit and that's all.
I thought it could use an update to make it feel it has more weight,
thus may I introduce new sounds on damage.

When certain types of damage occur a unique sound will play heard only on the interior.
This is also useful for those inside who can identify the sounds as they will 
know what kind of damage they just took.
These damage types are:
- Any explosion type damage
- Physical Collision under - 900 Velocity where damage would occur
- Physical Collision Over 900 Velocity
- On Death

Props to JEREDEK for providing the sounds used.


CrethFox:
wrong merge

PR #623, opened 2022-02-12
Random vortex position

### LINKED ISSUE #533: Fall out at a random place after undo-ing TARDIS in vortex

DoctorErt:
Closes #533

This should also fix the bugs where parts of the world are seen in the vortex

изображение


parar020100:
From Matt: (migrated from previous fork PR)
> Hey thanks for the contribution, so the code looks good but in-game the TARDIS suddenly jumps up into the sky before the vortex has time to fade in so it doesn't provide a smooth experience anymore as it did before. The vortex fades in over about 2 seconds clientside so to allow time for networking and lag I'd recommend adding a 3 second delay to this teleport.

lua/tardis/screens/cl_screen_coordinates.lua
```diff
@@ -138,28 +138,43 @@ TARDIS:AddScreen("Destination", {id="coordinates", text="Coordinates", menu=fals
...
+                        function()
+                            pos = ext:GetPos()
+                            ang = ext:GetAngles()
+                            if name == "" then name = "Unnamed" end
+                            TARDIS:AddLocation(pos,ang,name,map)
+                            updatelist()
+                        end,
+                        "No",
+                        function()
+                            TARDIS:AddLocation(pos,ang,name,map)
```
AmyJeanes, line 171: This seems to always be 0,0 no matter what, what's the intended behaviour here?

AmyJeanes (changes requested):

lua/tardis/screens/cl_screen_coordinates.lua
parar020100: That's a mistake, but an old one.
It behaves absolutely the same way on "release" branch

I've opened an issue for this --> #624, we'll try to fix that soon
However, since this PR isn't the reason for this bug, I suggest to merge it now and make the fix separately

parar020100: That's a mistake, but an old one.
It behaves absolutely the same way on "release" branch

I've opened an issue for this --> #624, we'll try to fix that soon
However, since this PR isn't the reason for this bug, I suggest to merge it now and make the fix separately

@MattJeanes FYI

AmyJeanes: Okay, sounds good I was thinking that would have also done this before too

AmyJeanes (approved):

PR #626, opened 2022-02-15
Mergeable texture sets

parar020100:
A small addition to everything added before, which simplifies setting up textures
Since that PR is close to being merged, I thought it might be more convenient to have these new changes separately

**P.S. if you merge custom settings first, please change base branch to "dev"**


lua/tardis/interiors/templates/ttcapsule.lua
```diff
@@ -35,7 +35,15 @@ T = {
                 angoffset=Angle(0,0,0)
             },
         }
-    }
+    },
+    Interior = {
```
AmyJeanes, line 39: This appears to have broken the default interior TT version with the interior door backfaces now being invisible
image

lua/entities/gmod_tardis/cl_init.lua
```diff
@@ -2,4 +2,5 @@ include('shared.lua')
 
 ENT:AddHook("PlayerInitialize", "interior", function(self)
     self.metadata=TARDIS:GetInterior(net.ReadString())
+    self.metadata = TARDIS:MergeInteriorTemplates(self.metadata, true, self:GetCreator(), self)
```
AmyJeanes: Could this not be done at the stage of TARDIS:GetInterior, perhaps by passing a couple extra parameters into that? Same for all other calls of this function

AmyJeanes (changes requested):
What's the best way to test this? I see the 1963 development branch has TextureSets but not really sure what I'm looking for other than it not crashing - how could I effectively test this?

lua/tardis/interiors/templates/ttcapsule.lua
parar020100: Fixed

parar020100:
**Testing texture sets**:
- Normal texture sets are used in season 10 ("next-core-version" branch), hellbent and 1963 ("dev" branches)
They control the textures with power off, power on and warning-state 
(you might like season 10 for this test, since it has a lot of dependent visual changes)

P.S. There's a `tardis2_debug_warning` concommand to switch the TARDIS between 10% and 100% health
I added it for extension devs who need to make the textures / lights for it, and you can use it while testing

- Texture sets that inherit each other are currently only in 1963
The use-case is not quite standard, since I have a template-generating function that create four texture sets in template format; then, they are applied to certain interiors

**Testing the conditional templates:**
- 1963 extension
"TT Capsule" is basically a conditional template

lua/entities/gmod_tardis/cl_init.lua
parar020100: Good idea
Done

AmyJeanes (approved):

PR #631, opened 2022-02-21
Added Prompt for Low Model Detail Settings

### LINKED ISSUE #475: Prompt when not using high model detail
The legacy provides a prompt when not using high model detail and offers to change it: https://github.com/MattJeanes/TARDIS/blob/legacy/lua/entities/sent_tardis/cl_init.lua#L345-L347

We should do this for the rewrite, providing an option for extension makers to allow them to show this prompt or not

I don't think we can automatically detect if this will happen for any particular model:

image

Artixc:
On Spawn Query,
image

Fixes #475

### DIFF
--- a/lua/entities/gmod_tardis_interior/cl_init.lua
+++ b/lua/entities/gmod_tardis_interior/cl_init.lua
@@ -7,8 +7,5 @@ ENT:AddHook("PlayerInitialize", "interior", function(self)
     self.Fallback=self.metadata.Interior.Fallback
     self.Portal=self.metadata.Interior.Portal
     self.ExitDistance=self.metadata.Interior.ExitDistance
-    if GetConVarNumber("r_rootlod")>0 then
-        Derma_Query("The TARDIS Interior may require model detail on high, set now?", "TARDIS Interior", "Yes", function() RunConsoleCommand("r_rootlod", 0) end, "No", function() end)
-    end
 end)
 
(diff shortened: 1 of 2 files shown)


lua/entities/gmod_tardis_interior/cl_init.lua
```diff
@@ -7,4 +7,8 @@ ENT:AddHook("PlayerInitialize", "interior", function(self)
     self.Fallback=self.metadata.Interior.Fallback
     self.Portal=self.metadata.Interior.Portal
     self.ExitDistance=self.metadata.Interior.ExitDistance
-end)
\ No newline at end of file
+    if GetConVarNumber("r_rootlod")>0 then
```
AmyJeanes: Can we move this out to a separate module? Something like `TARDIS/lua/entities/gmod_tardis_interior/modules/cl_lod_prompt.lua`

Also hooking on `PlayerEnter` may make a bit more sense so that we're only bothering people interacting with the TARDIS in a multiplayer scenario rather than everyone.

Can we add a check as well to see if there is an optional thing set on the metadata, maybe `self.metadata.Interior.RequireHighModelDetail`? If it is not set (nil) then assume that it is required. This will allow extension developers to disable the prompt if not required for their interior.

AmyJeanes (changes requested):

AmyJeanes (approved):
Looking great! Congratulations on your first PR for the TARDIS!

PR #633, opened 2022-02-21
Added Custom Lighting Prompt, Reformatted lod prompt

Artixc:
Tardis Interior Developers may now put
    RequireLightOverride = true,
in T.Interior to request the user to turn on experimental lighting overrite.
image


lua/entities/gmod_tardis_interior/modules/cl_customlighting_prompt.lua
```diff
@@ -0,0 +1,29 @@
+--Require Lighting Override
```
AmyJeanes: I think since this is similar enough to the other one we can actually put them both into the same file. GMod has limits on how many Lua files can be loaded so I like to reduce them when possible. Can we merge the two files together into like `cl_prompts.lua`?

lua/entities/gmod_tardis_interior/modules/cl_customlighting_prompt.lua
```diff
@@ -0,0 +1,29 @@
...
+    type="bool",
+})
+
+ENT:AddHook("PlayerInitialize", "customlightenable", function(self)
+    local require_override = (self.metadata.Interior.RequireLightOverride == true) 
+    local light_override = TARDIS:GetSetting("lightoverride-enabled", false)
+    local noshow = TARDIS:GetSetting("light_override_request_noshow", false)
+    if require_override and not light_override and not noshow then
+        Derma_Query(
+            "It is recommended to use Lighting Override with this TARDIS! This can be changed in the TARDIS settings later. Enable now?",
```
AmyJeanes: The exclamation mark is a little jarring in this message, how about this?

```suggestion
            "This TARDIS recommends Lighting Override which you do not have enabled. Enable now? This can be changed in TARDIS settings later.",
```

lua/entities/gmod_tardis_interior/modules/cl_customlighting_prompt.lua
```diff
@@ -0,0 +1,29 @@
...
+            "TARDIS Interior",
+            "Yes",
+            function()
+                TARDIS:SetSetting("lightoverride-enabled", true)
+            end,
+            "No",
+            nil,
+            "Dont Show Again",function()
```
AmyJeanes: Nice! I think people will definitely appreciate this.

AmyJeanes (changes requested):

lua/entities/gmod_tardis_interior/modules/cl_customlighting_prompt.lua
Artixc: I hope so too!

AmyJeanes (approved):
Looks great, nice work!

PR #639, opened 2022-03-12
HOTFIX: GetData returning false fixed

parar020100:
This is a potentially very serious bug - GetData() wasn't able to return "false" 

P.S. If you agree to make a release for this fix, this PR should be closed


AmyJeanes:
Will merge directly from release

PR #642, opened 2022-03-14
Major settings update

### LINKED ISSUE #436: Have all TARDIS settings in the right side of the spawnmenu (options)
"Settings" screen has proven to be very ineffective and inconvenient.

### LINKED ISSUE #548: GUI Rows number is not saved when configured as a number
From Blazi on gitlab:

When selecting what number of rows a user would like to have for the new GUI. you cannot save it by editing the number in the bar.
The slider has to be used or it wont save and update the GUI.

### LINKED ISSUE #629: Unite sound categories
изображение

parar020100:
Closes #436, #548, #629

- Options are now in the right side of spawnmenu
  - Automatically-generated list of settings
  - More features like subsections and "reset section settings"
- Refactored settings module
  - added `setting.class` to define networking behaviour
    - setting.class is either `local`, `networked` or `global`
  - function `GetSetting()` can now have both player and the TARDIS entity as a parameter
    - if a non-player entity is specified, the setting is taken from its creator
  - cleanup
    - Purged `default` parameter of `GetSetting()`
    - Purged `networked` and `broadcast` parameters of `SetSetting()`
    - Purged `networked` parameter of the setting data
    - Purged `settings` control
  - More safeguards ensuring the correct setting behaviour 
  - Fixed setting types, "list" and "integer" types support
  - `round_func` function of the setting to make sure its round value is selected
- Added global settings to the options
  - Added automatic links of the global settings to convars
- Settings are now defined in separate folder (`lua/tardis/settings/`) 
- Reorganised setting categories
- Removed settings from the GUI
  - Forbade to open GUI from outside the TARDIS

- TARDIS Data (SetData/GetData) is all stored in the exterior 
  - Interior functions are redirected to the exterior
- Added `CallCommonHook()` function which calls the hook both for the exterior and interior
- Added support for metadata hooks for both exterior and interior hooks

- Extended the range of music volume setting (1..100 -> 1..1000)
- Renamed "Use VisGUI" into "Use old GUI" with inverted values
- Lighting Override is on by default and no longer is considered experimental
- New HUD TARDIS Icon
- Moved tardisdebug-related code into a separate file

- Bug fix: interrupted teleport sounds used to be handled serverside


lua/entities/gmod_tardis_interior/modules/libraries/sh_data.lua
```diff
@@ -1,80 +1,25 @@
...
-                net.Send(ply)
-            else
-                net.Broadcast()
-            end
-        end
+        self.exterior:SendData(ply)
     end
-    
+
     hook.Add("PlayerInitialSpawn", "TARDISI-Data", function(ply)
```
AmyJeanes: This hook likely isn't required if it's sourcing from the exterior only now

lua/tardis/cl_menuoptions.lua
```diff
@@ -0,0 +1,201 @@
+hook.Add("PopulateToolMenu", "TARDIS2-PopulateToolMenu", function()
+    -- Options
+    local options={}
+    local sections={}
+    local subsections={}
+    for k,v in pairs(TARDIS:GetSettingsData()) do
+        if v.option then
+            table.insert(options,{v.id, v, (v.subsection or " ") .. v.name})
```
AmyJeanes, line 8: What's the string with a single space for here?

lua/tardis/cl_menuoptions.lua
```diff
@@ -0,0 +1,201 @@
...
+            end
+        end
+    end
+
+    table.SortByMember(options, 3, true)
+    table.SortByMember(sections, 1, true)
+
+    for k,section in ipairs(sections) do
+        local section_id = "TARDIS2_Options_" .. section
+        local section_text = " " .. section
```
AmyJeanes, line 24: Why are we prefixing sections with a space? This is inconsistent vs other options in the spawnmenu and a hack 
image

lua/tardis/cl_menuoptions.lua
```diff
@@ -0,0 +1,201 @@
...
+            local reset_button = vgui.Create("DButton")
+            reset_button:SetText("Reset settings (this section)")
+            reset_button.DoClick = function(self)
+                Derma_Query(
+                    "Reset clientside settings of section \"" .. section .. "\"? This cannot be undone.",
+                    "TARDIS Interface",
+                    "OK", function()
+                        TARDIS:ResetSectionSettings(section)
+                        TARDIS:Message(LocalPlayer(), "TARDIS clientside settings of section \"" .. section ..  "\" have been reset. You may need to respawn the TARDIS for all changes to apply.")
+                        RunConsoleCommand("spawnmenu_reload")
```
AmyJeanes: This works but is a bit hacky, is there a better approach like reloading the controls of this section only?

lua/tardis/cl_menuoptions.lua
```diff
@@ -0,0 +1,201 @@
...
+        local button = vgui.Create("DButton")
+        button:SetText("Reset clientside settings")
+        button.DoClick = function(button)
+            Derma_Query(
+                "Reset all clientside settings? This cannot be undone.",
+                "TARDIS Interface",
+                "OK", function()
+                    TARDIS:ResetSettings()
+                    TARDIS:Message(LocalPlayer(), "TARDIS clientside settings have been reset. You may need to respawn the TARDIS for all changes to apply.")
+                    RunConsoleCommand("spawnmenu_reload")
```
AmyJeanes, line 204: Ditto on hacky method

lua/tardis/cl_options.lua
```diff
@@ -96,70 +96,126 @@ function TARDIS:ChangeOption(id,data)
...
+        tooltip = tooltip .. "\n\nDefault: " .. tostring(data.value)
+    end
+
+    local elem
+    local elem2 = nil
+
+    local setting = TARDIS:GetSetting(id)
+
+    local function AreColorsDifferent(a,b)
+        return (a.r ~= b.r) or (a.g ~= b.g) or (a.b ~= b.b) or (a.a ~= b.a)
```
AmyJeanes: This appears unused?

lua/tardis/cl_options.lua
```diff
@@ -96,70 +96,126 @@ function TARDIS:ChangeOption(id,data)
...
-        if v.option then
-            table.insert(options,{v.name,v.id,v})
-            if v.section and not table.HasValue(sections,v.section) then
-                table.insert(sections,v.section)
+function TARDIS:CreateOptionInterface(id, data)
+
+    local text = data.name or id
+    local tooltip = data.desc or ""
+    if data.value ~= nil then
+        tooltip = tooltip .. "\n\nDefault: " .. tostring(data.value)
```
AmyJeanes, line 8: Tooltip for lists results in the raw value 
image

Other types seem good

lua/tardis/interiors/modules/sh_customsettings.lua
```diff
@@ -1,17 +1,9 @@
...
-    id = "interior_custom_settings",
-    name = "Interior-specific settings",
-    value = {},
-    option = false,
-    networked = true
-})
-
 function TARDIS:GetCustomSettings(ply)
-    return TARDIS:GetSetting("interior_custom_settings", {}, ply) or {} -- nil is never allowed
+    return TARDIS:GetSetting("interior_custom_settings", ply) or {} -- nil is never allowed
```
AmyJeanes, line 2: Does the new default system handle this scenario, so we can move that empty table fallback there?

lua/tardis/libraries/sh_settings.lua
```diff
@@ -80,66 +49,236 @@ function TARDIS:GetSettingsData()
     return self.SettingsData
 end
 
-function TARDIS:SetSetting(id,value,networked,broadcast)
+--------------------------------------------------------------------------------
+-- Accessing
+
+function TARDIS:SetSetting(id, value)
+    local data = self.SettingsData[id]
+    if not data then error("Requested setting ", id, " does not exist") end
```
AmyJeanes: Did you mean to use concat here?

```suggestion
    if not data then error("Requested setting ".. id  .." does not exist") end
```

lua/tardis/libraries/sh_settings.lua
```diff
@@ -80,66 +49,236 @@ function TARDIS:GetSettingsData()
...
-        return self.Settings[id]
-    elseif CLIENT and (self.LocalSettings[id] ~= nil) then
-        return self.LocalSettings[id]
-    elseif CLIENT and (self.NetworkedSettings[id]~=nil) then
-        return self.NetworkedSettings[id]
-    else
-        return default
+function TARDIS:GetSetting(id, src)
+    local ply
+    if IsEntity(src) and src:IsPlayer() then
```
AmyJeanes: This can be slightly optimised to only call IsEntity once e.g.

```lua
if IsEntity(src) then
  if src:IsPlayer() then
     ...
   else
     ...
   end
end
```

This could affect performance as settings can be called every frame multiple times

lua/tardis/libraries/sh_settings.lua
```diff
@@ -80,66 +49,236 @@ function TARDIS:GetSettingsData()
...
+            if self.ClientSettings[ply:UserID()] then
+                return select_return_val(self.ClientSettings[ply:UserID()][id])
+            end
+            return select_return_val(nil)
+        end
+
+        error("Networked setting " .. id .. " was requested for invalid player " .. tostring(ply))
     end
+
+    error("Requested setting " .. "either doesn't exist or has no defined class")
```
AmyJeanes: Mean to put the id in here?

```suggestion
    error("Requested setting " .. id .. "either doesn't exist or has no defined class")
```

lua/tardis/libraries/sh_settings.lua
```diff
@@ -80,66 +49,236 @@ function TARDIS:GetSettingsData()
...
 
-function TARDIS:SendSetting(id,value,ply)
-    net.Start(SERVER and "TARDIS-Settings" or "TARDIS-ClientSettings")
-        net.WriteBool(true)
-        net.WriteType(id)
-        net.WriteType(value)
+--------------------------------------------------------------------------------
+-- Saving
+
+local LOCAL_SETTINGS_FILE = "tardis_settings_cl.txt"
```
AmyJeanes, line 171: Are these settings compatible with the previous version or is it a start fresh situation?

lua/tardis/libraries/sh_settings.lua
```diff
@@ -80,66 +49,236 @@ function TARDIS:GetSettingsData()
...
+end
+
+function TARDIS:LoadSettings()
+
+    local function LoadSettingsFromFile(settings_table, settings_file)
+        if file.Exists(settings_file, "DATA") then
+            local file_contents = file.Read(settings_file, "DATA")
+            local loaded_settings = self.von.deserialize(file_contents)
+
+            table.Merge(settings_table, loaded_settings)
```
AmyJeanes, line 203: How does this work if we are converting the settings as key value above when merging back in for load?

AmyJeanes (changes requested):
It's looking really really good, nice work! Other than a couple of minor points / questions I'm happy with this and in-game it's functioning great!

lua/tardis/cl_options.lua
```diff
@@ -96,70 +96,126 @@ function TARDIS:ChangeOption(id,data)
...
+        spacer:SetTall(2)
+        elem:AddItem(spacer)
+
+    elseif data.type=="list" then
+        elem = vgui.Create("DLabel")
+
+        elem2 = vgui.Create("DComboBox")
+
+        if data.get_values_func ~= nil then
+            for k,v in pairs(data.get_values_func()) do
```
AmyJeanes, line 107: Can this not be a direct k,v mapping here to optimise lookup + performance? Right now we seem to only be using the values and ignoring the key

lua/tardis/cl_menuoptions.lua
parar020100: As in spawnmenu's custom settings, there are two types of spaces
The point of them is sorting the options
"Reset All Settings" needs to go to the bottom

lua/tardis/cl_menuoptions.lua
parar020100: Making the normal options go before all the subsections

lua/entities/gmod_tardis_interior/modules/libraries/sh_data.lua
parar020100: Fixed

lua/tardis/cl_options.lua
parar020100: Removed

lua/tardis/libraries/sh_settings.lua
parar020100: Sure, fixed

lua/tardis/libraries/sh_settings.lua
parar020100: Added, thanks

lua/tardis/libraries/sh_settings.lua
parar020100: Yes, fixed

lua/tardis/libraries/sh_settings.lua
parar020100: 100% compatible
File names are the same

lua/tardis/libraries/sh_settings.lua
parar020100: Honestly, I have no idea
I haven't changed the way they are saved or loaded - just refactored the process

lua/tardis/cl_options.lua
parar020100: https://wiki.facepunch.com/gmod/DComboBox
There doesn't seem to be an `AddChoices()` function (or similar)
So we have to use AddChoice for each choice separately and have that extra loop there

lua/tardis/cl_menuoptions.lua
AmyJeanes: Do they not get added to the menu in the order they are added?

lua/tardis/libraries/sh_settings.lua
AmyJeanes: Well it seems to work so whatever I guess lol

lua/tardis/cl_menuoptions.lua
AmyJeanes: Is it visible to the user?

parar020100: Yes
If you remove it, it looks all over the place = horrible

lua/tardis/cl_menuoptions.lua
parar020100: Not sure, but they used to be sorted at some point
Also, sometimes the menus change order when you mess with the code
IMO, it's better to have control over what is where, since the list is generated automatically anyway

lua/tardis/cl_options.lua
parar020100: изображение
Fixed

lua/tardis/interiors/modules/sh_customsettings.lua
parar020100: It is handled and actually should never be nil, in this case, {} is there just in case

lua/tardis/cl_menuoptions.lua
AmyJeanes: Okay, it's not ideal but I get not something we have a lot of control over. Can always revisit this later if a better solution comes up

lua/tardis/cl_menuoptions.lua
AmyJeanes: Okay, again not super ideal but it'll do for now

lua/tardis/cl_menuoptions.lua
AmyJeanes: Now resolved, thanks!

lua/tardis/cl_menuoptions.lua
AmyJeanes: Also now resolved!

lua/tardis/cl_options.lua
AmyJeanes: Do we need to show the number here as well? Just showing 'Default: In-built notifications' would be perfect imo

lua/tardis/interiors/modules/sh_customsettings.lua
AmyJeanes: 👍

lua/tardis/cl_menuoptions.lua
parar020100: Done
I've made the section settings refresh without spawnmenu reload

I'm keeping the hacky method for the global refresh though, since it's much more complicated
You can do it yourself if you want, but I don't think it's worth the effort

lua/tardis/cl_menuoptions.lua
parar020100: Replied above

lua/tardis/cl_options.lua
parar020100: It's value + text in all such things
Values are not required here, but might be useful in other situations
Plus, it was easier to add (I just added the text to existing tip, instead of making the tip conditional

parar020100: How much do you want this to change?
(tbh, I'm lazy to waste too much time on perfectionism noone would even notice)

AmyJeanes: Not that bothered was mostly just curious, works for me!

AmyJeanes (approved):

PR #648, opened 2022-04-11
Fixes for multiple errors and problems

### LINKED ISSUE #645: Errors to investigate
I don't think I remember the sources and conditions for all of these
Before the next release, they need to be checked out / investigated / looked into

1:
изображение

2:
изображение

3:
изображение

parar020100:
- Closes #645
- Adds some functions for extension makers
- Fixed the vortex glitching
  The only way I was able to do that is removing the timer
  It looks a bit weird when you get into the vortex from a random point, but still much more smooth than vortex disappearing


lua/tardis/sh_parts.lua
```diff
@@ -28,10 +28,14 @@ function TARDIS.DrawOverride(self,override)
             if self.PreDraw then self:PreDraw() end
             if self.UseTransparencyFix and (not override) then
                 render.SetBlend(0)
-                self.o.Draw(self)
+                if self.o.Draw ~= TARDIS.DrawOverride then
```
AmyJeanes: Did you see if this was needed after what we discussed? It shouldn't be possible to get into a state where this can happen

lua/entities/gmod_tardis/modules/teleport/sh_tp_main.lua
```diff
@@ -307,9 +307,7 @@ if SERVER then
             and not self:GetData("redecorate")
             and not self:GetData("redecorate_parent")
         then
-            self:Timer("VortexChangePositionTime", 3, function()
-                self:ChangePosition(self:GetRandomLocation(false), self:GetAngles(), false)
-            end)
+            self:ChangePosition(self:GetRandomLocation(false), self:GetAngles(), false)
```
AmyJeanes, line 310: What issues were there with the timer? I'd much rather fix those if possible it's a pretty jarring jump as soon as you finish dematting. We maybe just need to ensure that we are still in the correct state (dematted, in vortex, not matting, tardis is valid etc)

AmyJeanes (commented):
Looking pretty good as far as I can tell, couple comments on some behaviour and also I found that the added validation on TARDIS:Message broke the ALT+E hint as that was passing in a nil player, so fixed that - as far as I can tell all the other calls to TARDIS:Message correctly pass in the player.

lua/entities/gmod_tardis/modules/teleport/sh_tp_main.lua
parar020100: The vortex just glitches out
When it is moved and the TARDIS is moved, the vortex is invisible until you start moving your camera around
It happens often, and looks horrible

lua/tardis/sh_parts.lua
parar020100: We did discuss that it shouldn't be possible
But it still happened with that "adding parts in the same file" situation
We might add an error message there if you want, but I want to keep this to prevent stack overflow. Just as a safeguard

AmyJeanes: Yeah I get that but my issue is that these checks will be running for every part potentially multiple times per frame on every TARDIS and even a simple if statement can impact performance here. If we want to add safeguards I would add them into the Initialize or SetupOverrides if possible

lua/entities/gmod_tardis/modules/teleport/sh_tp_main.lua
AmyJeanes: I've seen this happen before with the standard vortex as well, but I agree I have seen this happening as well. I think given the problem that it currently has this is a better solution for now, but I'd like to log it as an issue to be investigated but I have a suspicion that it's going to be down to limitations in rendering.

lua/tardis/sh_parts.lua
parar020100: Reverted them for now
Created issue #650

AmyJeanes (approved):

PR #672, opened 2022-07-15
add Chinese translations

Fox2shadow:


AmyJeanes:
Hi @ALexZekio thanks for the contribution

This looks great - well I have no idea if the translations are correct but I'll take your word for it

There is one small issue in that all files in GMod need to be lowercase so could I request that you rename the files to be lowercase such as `zh-cn.lua` and `zh-hk.lua`

Many thanks

PR #673, opened 2022-07-15
add Chinese translations

Fox2shadow:


AmyJeanes:
Hi again, thanks for the changes - by the way you don't need to close the pull request and re-open it to make changes you can just make the change and it will update automatically 😄 

Chinese Simplified is working, looks good!
image

I have noticed that the Chinese Traditional language code in GMod is `zh-TW` and not `zh-HK` as you have entered so could I ask you please rename the file and the T.Code for this and then it should be all good!

You can see a list of supported language codes in GMod here: https://wiki.facepunch.com/gmod/Addon_Localization#supportedlanguages

Fox2shadow:
I have changed zh-HK to zh-TW, Thanks for reminding. :)

AmyJeanes (approved):
Thanks - all working now!

Please note that I am not yet done adding translations to the `en` file so you might want to re-visit these when that is updated - I will let you know when they are

AmyJeanes:
Also please note that this will not be deployed to Steam Workshop until all of the translations are finished 🙂

PR #674, opened 2022-07-26
Add more Chinese translations

Fox2shadow:
Follow up the simplified Chinese translation. :)

### DIFF
--- a/lua/tardis/languages/zh-cn.lua
+++ b/lua/tardis/languages/zh-cn.lua
@@ -5,8 +5,25 @@ T.Code = "zh-CN"
 T.Name = "简体中文"
 T.Phrases = {
     ["Common.Error"] = "错误",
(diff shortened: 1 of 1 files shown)


AmyJeanes:
Wow this is great, thank you! There are still more translations to add to the English base language but I'm slowly getting through them :)

I noticed some of the ones are still in English - is that intentional? If you leave any of them out they will automatically fall back to English anyway so you don't need to specify them

Fox2shadow:
I have fixel the missing part, Thanks for reminding me. :P

AmyJeanes (approved):
Looks good in-game! I have no idea what they mean so I'll have to take your word for it that they are sensible translations 😄 

image
image

As I mentioned there is still more to go and I will let you know when it's ready

PR #675, opened 2022-07-27
Warning features

CrethFox:
Added support for color change to exterior lamp light when changing to warning mode.
Added support for warning mode sequence speeds.


lua/tardis/interiors/default.lua
```diff
@@ -223,6 +223,11 @@ T.Exterior = {
     Parts = {
         door = true
     },
+    Light = {
+        enabled = true,
```
AmyJeanes: You can get rid of the enabled/pos here as they are inherited from the base interior

AmyJeanes (approved):

PR #696, opened 2022-08-27
Remat sound instead of damaged mat sound

PoogieTheGoose:


lua/entities/gmod_tardis/modules/teleport/sh_tp_main.lua
```diff
@@ -369,9 +369,9 @@ else
                 if pos and self:GetData("demat-fast",false) then
                     if not IsValid(self) then return end
                     if (self:GetData("demat-fast",false))==true then
-                        sound.Play(ext.mat_damaged, pos)
+                        sound.Play(ext.mat_fast, pos)
                     else
-                        sound.Play(ext.mat, pos)
```
AmyJeanes, line 374: Is it even possible to hit this code? We check for demat-fast on line 369 and then again on 371

Probably can just delete that 2nd if statement unless I'm missing something

lua/tardis/interiors/base.lua
```diff
@@ -91,6 +91,7 @@ T.Exterior = {
             mat = "drmatt/tardis/mat.wav",
             mat_damaged = "jeredek/tardis/mat_damaged.wav",
             mat_fail = "drmatt/tardis/mat_fail.wav",
+            mat_fast = "Poogie/tardis/mat_fast.wav",
```
AmyJeanes: All files must be lowercase for Linux/macOS compatibility

Note renaming the actual file often won't get picked up you may have to change it to something else entirely, commit, then change it back with lowercase and commit again

AmyJeanes (changes requested):

lua/entities/gmod_tardis/modules/teleport/sh_tp_main.lua
AmyJeanes: Looking further down it seems the 2nd if statement should instead be checking for `self:GetData("health-warning", false)` like it does on line 407, maybe it should be doing that. Do we have a fast mat sound for when it's damaged?

PoogieTheGoose: > Looking further down it seems the 2nd if statement should instead be checking for `self:GetData("health-warning", false)` like it does on line 407, maybe it should be doing that. Do we have a fast mat sound for when it's damaged?

No but i can make one.

AmyJeanes: Cool okay that'd be nice. I assume we need the fast mat sound to ensure it all syncs up nicely with the visuals when doing a fast demat and can't just use the regular mat sound?

PoogieTheGoose: > Looking further down it seems the 2nd if statement should instead be checking for `self:GetData("health-warning", false)` like it does on line 407, maybe it should be doing that. Do we have a fast mat sound for when it's damaged?

It's not health warning. I tried to change the sound and it was the same as 1st in warning state

PoogieTheGoose: > > Looking further down it seems the 2nd if statement should instead be checking for `self:GetData("health-warning", false)` like it does on line 407, maybe it should be doing that. Do we have a fast mat sound for when it's damaged?
> 
> It's not health warning. I tried to change the sound and it was the same as 1st in warning state

It seems like it's actually doesnt have any use

AmyJeanes: Is that with that code change in place? Right now it seems useless but if you change the if condition it might start doing something

PoogieTheGoose: > Is that with that code change in place? Right now it seems useless but if you change the if condition it might start doing something

Done

lua/tardis/interiors/base.lua
```diff
@@ -91,6 +91,8 @@ T.Exterior = {
             mat = "drmatt/tardis/mat.wav",
             mat_damaged = "jeredek/tardis/mat_damaged.wav",
             mat_fail = "drmatt/tardis/mat_fail.wav",
+            mat_fast = "p00gie/tardis/mat_fast.wav",
+            matdamaged_fast = "p00gie/tardis/matdamaged_fast.wav",
```
AmyJeanes: Could we make this `mat_damaged_fast` to be consistent with `mat_damage`?

AmyJeanes: Also please can you change the actual sound filename the same way!

PoogieTheGoose: > Could we make this `mat_damaged_fast` to be consistent with `mat_damage`?

If you mean the sound itself to be consistent with demat effect i will do it soon(im working on my gmod video right now)

AmyJeanes: No I mean the name of the variable in the code and the filename of the sound

PoogieTheGoose: > No I mean the name of the variable in the code and the filename of the sound

I did both

AmyJeanes:
Perfect thanks! Code all looks good I'll give it a go in game and check out the sounds later

AmyJeanes (approved):
Okay all seems to function as expected although it would be nice to have the sounds have the incoming materialisation effect like the normal one. You mentioned that you were working on this soon? Not sure if you plan to include that as part of this PR or a different one - let me know what you plan to do but otherwise I'm happy with this!

PoogieTheGoose:
> Okay all seems to function as expected although it would be nice to have the sounds have the incoming materialisation effect like the normal one. You mentioned that you were working on this soon? Not sure if you plan to include that as part of this PR or a different one - let me know what you plan to do but otherwise I'm happy with this!

I just planned a small change for the sounds because it was just uncomfortable to hear broken mat sound on the exterior so i changed it to something different. I really liked the idea of mat quickly without incoming sounds like it was in the show
[Fast Mats](https://youtu.be/E4i3_rqe8uc)

PoogieTheGoose:
> > Okay all seems to function as expected although it would be nice to have the sounds have the incoming materialisation effect like the normal one. You mentioned that you were working on this soon? Not sure if you plan to include that as part of this PR or a different one - let me know what you plan to do but otherwise I'm happy with this!
> 
> I just planned a small change for the sounds because it was just uncomfortable to hear broken mat sound on the exterior so i changed it to something different. I really liked the idea of mat quickly without incoming sounds like it was in the show [Fast Mats](https://youtu.be/E4i3_rqe8uc)

So thats all i wanted to change.Thanks for reviewing!

PR #712, opened 2022-11-12
Made interrupting teleport more smooth

parar020100:
* Fixed: Interrupting teleport doesn't turn off the flight mode and floating any more

* Tweaked: You can use the handbrake while dematerialising to cancel it, and the TARDIS will not explode
It will explode if pressed in the vortex or while materialising
That is screen-accurate with Clara using the handbrake in "Kill the Moon"


AmyJeanes (approved):
Looks good and working as described in-game!

PR #720, opened 2022-11-19
Light states feature, dynamic lights optimizations and more

parar020100:
- Lots of light optimizations (FPS in interiors with 3 lights increased by a lot)
- Lots of refactoring
- normal, off, warning, off_warning states support for each light or lamp
- Added light states (which apply both to lights and lamps)
    - creators can specify light states for each light and lamp separately
    - when a light state is applied, the TARDIS looks through all of its light sources and checks whether any of them have the light state with that ID; if so, it is applied
- New way of specifying lights & lamps metadata (see below)
    - previous ways still supported


parar020100:
An example of light metadata in the new format:
```lua
{
    color = Color(255, 50, 0),
    pos = Vector(0, 0, 120),
    brightness = 5,
    falloff = 10,

    warn_color = Color(255, 0, 0),
    warn_pos = Vector(0, 0, 120),
    warn_brightness = 3,
    warn_falloff = 10,

    nopower = true,

    off_color = Color(0, 0, 160),
    off_pos = Vector(0, 0, 80),
    off_brightness = 1,
    off_falloff = 10,

    off_warn_color = Color(0, 0, 160),
    off_warn_pos = Vector(0, 0, 80),
    off_warn_brightness = 1,
    off_warn_falloff = 10,

    states = {
        lstate_one_name = {
            -- each state can have the same set of parameters as the default one
            color = Color(234, 255, 0),
            brightness = 3,
            falloff = 20,
        
            warn_color = Color(255, 10, 0),
            warn_pos = Vector(0, 0, 130),
            warn_brightness = 2,
            warn_falloff = 20,
        
            off_pos = Vector(0, 0, 80),
        
            off_warn_color = Color(0, 0, 160),
            off_warn_pos = Vector(0, 0, 80),
            off_warn_brightness = 1,
        }, 
        disabled = {
            -- applying this state would stop this light from rendering
            enabled = false,
        },
        nopower_full_darkness = {
            nopower = false,
            -- applying this light state would stop lights from appearing with no power
        }
    },
},
```

All options are inheritable, you don't have to specify the ones you don't need to change
`warn_*` parameters inherit the default ones
if `nopower` option is true, `off_*` params inherit the default ones
if `nopower` option is true, `off_warn_*` params inherit the `off_*` ones 

The `warncolor` parameter is still supported, though deprecated
It is recommended to use `warn_color`

parar020100:
An example of lamp metadata in the new format:

```lua
lamp_id = {
	color = Color(0, 255, 230),
	texture = "effects/flashlight/soft",
	fov = 169.09091186523,
	distance = 165.45454406738,
	brightness = 0.3,
	pos = Vector(329.08, -88.87, 74.012),
	ang = Angle(87.64, 20.88, 169.94),

	nopower = true,
	
	warn = {
	    -- same list of options available
	    color = Color(200, 255, 230),
            brightness = 2.3,
            texture = "effects/flashlight/soft",
	    fov = 169.09091186523,
	    distance = 165.45454,
	    pos = Vector(329.08, -88.87, 74.012),
	    ang = Angle(87.64, 20.88, 169.94),
	},
	off = {
	    -- same list of options available
	    color = ...
	    brightness = ...
	    pos = ...
	    ang = ...
            texture = ...
            fov = ...
	    distance = ...
	},
	off_warn = {
	    -- same list of options available
      	    color = ...
	    brightness = ...
	    pos = ...
	    ang = ...
	},
	
	states = {
	    state_id_1 = {
	        color = ...
	        brightness = ...
	        pos = ...
	        ang = ...
	        ...
	        
	        warn = {...},
	        off = {...},
	        off_warn = {...},
	    },
	    state_id_2 = {...},
	    ...
	},
},
```

lua/entities/gmod_tardis_interior/modules/sh_interior_lights.lua
```diff
@@ -61,144 +161,270 @@ if CLIENT then
         end
     end)
 
-    function ENT:AddRoundThing(pos)
-        self.roundthings[pos]=util.GetPixelVisibleHandle()
-    end
+    ENT:AddHook("ShouldDrawLight", "enabled_check", function(self,id,light)
```
AmyJeanes: Hook should be named after the file it's in (and something else if there's multiple of the same hook in the same file)

lua/tardis/settings/cl_performance.lua
```diff
@@ -13,15 +13,27 @@ TARDIS:AddSetting({
 })
 
 TARDIS:AddSetting({
-    id="projlights-enabled",
+    id="extra-lights-disabled",
```
AmyJeanes: Can we inverse this option? Currently all the performance options require unticking to turn features off so this one is the odd one out 

image

AmyJeanes (changes requested):
Looking really nice, really great work! Just a couple of small things before we get this in 😄

lua/tardis/settings/cl_performance.lua
parar020100: Fixed

lua/entities/gmod_tardis_interior/modules/sh_interior_lights.lua
parar020100: Should be better now

AmyJeanes (approved):

PR #721, opened 2022-11-22
Old mat and demat drum beat

PoogieTheGoose:
So....
I changed mat,demat and full sounds with replaced drums in them(2010 drum beat ---> 1963-2005 drum beat).The reason for it is that many TARDIS extensions use base addon sounds and most of them are older interior ones(1983, etc) and it would be more easer to have two sound variants in the base addon(default being the old drum)so the old TARDISes will have it as default and the ones who want 2010 drum will just have to change the sound file path.It would be really nice if you accept those sounds and fade sequence for them(to make them more smoother for the sounds).
P.s:
I also changed mat fail sound, so it wont have this ugly vorp sound in the end


parar020100:
I'll listen to the sounds later
But anyway, I think the default interior's sounds need to stay
If you're changing the ones in base.lua, you need to restore the existing ones in default.lua

PoogieTheGoose:
> I'll listen to the sounds later But anyway, I think the default interior's sounds need to stay If you're changing the ones in base.lua, you need to restore the existing ones in default.lua

Done

parar020100:
Demat: all ok

image

parar020100:
Full flight: ok

image

parar020100:
Mat: 0.5 seconds offset (?)

image

parar020100:
mat_fail: has been made much better (with that junk lost at the end)

image

parar020100:
mat_fast: ok

image

parar020100:
mat_fast_damaged: ok

image

parar020100:
Mat has been fixed
I've double-checked the sounds in the game
All works fine for me, can be merged

AmyJeanes (approved):

Issue #176, opened 2016-09-05
In-game tutorial


MattJeanes:
It would be nice to have some sort of in-game tutorial similar to how GMod has it's own 'tips' that you can disable - this would be linked to #30.

So for example when you first enter the console it would pop up with a little hint saying to press E on the console or when you first enter third person to press R to fly or middle mouse to open the screen and things like that.

I've seen a few YouTubers mess with the rewrite and legacy and they seem to be missing a lot of features because they're not very well 'advertised'. Not sure about this feature though but it would be nice to have.

Noname12345GER:
Make sure LegacyHint isn't spammed, and one at a time.

Noname12345GER:
I would be for an simple Chat Print like, 

if you need help with the TARDIS, please say tardishelp

I thought of an Holo Tutorial NPC but translucency is our enemy there.

parar020100:
There are control tips now
I don't think this will be useful with those
We could make a short "help" text in the right panel or at addon's steam page
Or we could make a youtube video (or, for example, ask Septex to do it)

MattJeanes:
Control tips do help but they are using in-universe language and therefore do not easily explain what their actual functions are. I think maybe a button in the GUI for help which links to some documentation / YouTube video would be a good way to help this without loads of effort.

PR #722, opened 2022-11-22
Artron energy suggestions 

### LINKED ISSUE #42: Fuel system (Artron energy)
Cookiemorph's E2 had something like this. Not sure if I want to do something officially for it.

### LINKED ISSUE #651: Disable cloak when power is off

### LINKED ISSUE #654: cl_drawhud 0 doesn't disable TARDIS health indicator

parar020100:
- Artron energy
  - Closes #42 
  - A fuel system for the TARDIS
  - All numbers / values here have been experimentally balanced
  - The max. amount is configurable (1800 - 57600, 14400 by default) 
  - The initial amount is configurable (0/25/50/75/100%)
  
  - Recharges over time
    - +15/s in normal mode
    - very fast if handbrake is engaged (+180/s)
    - rather fast when power is off with no handbrake (+24/s)
    - slower with warning (+10/s)
    - much slower while floating (+8/s)
    - handbrake + float make it better than default but not too fast (+40/s)
   
  - If the TARDIS is charging, energy is added every 5 seconds
    If energy is wasted, it happens every 1 second
   
  - Is wasted on:
    - demat (-540)
    - mat (-360)
    - full flight (-1200)
    - vortex (-32/s)
    - flight (-45/s)
    - boost in flight (additional -45/s, making it -90/s in total)
    - cloak (-36/s)
      - much slower when handbrake is on (-32/s)
    - hads demat (-180)
      - energy isn't wasted on vortex time and mat after hads is triggered
    - failed demat (-80)
    - using some controls (single-time decrease)
      - toggle cloak (-720)
      - toggle physlock (-540)
      - toggle float (-540)

  - Doesn't stop repair or redecoration
    
  - When depleted:
    - If the TARDIS is teleporting, simply interrupts teleport
    - Otherwise, disables power
    - Possibly enables anti-gravs if flight was enabled
      - defined by a setting (see below), disabled by default

  - When too low:
    - Causes failed demat when attempted
    - Prevents HADS from triggering (if VERY low)
    - In vortex, causes failed mat

  - When equals 0:
    - Makes impossible to enable power

- New engine release function: "emergency charging"
  - Time vortex is filled with artron energy, but it's rather hard to charge it
  - When used in the vortex, engine release adds some artron energy (+720)
    - it also wastes quite a lot of TARDIS health (15-40% of current health randomly)

- An artron energy HUD indicator (in %)
  - is shown next to health
  - fonts have been tweaked to make them take less screen space (requested many times)
  - small refactoring of cl_hud.lua that simplified creating HUD options in the future
  - health and artron are independent, if one is disabled, another will move to be in the corner
 

- New settings:
  - Admin / Enable long repair
      - on by default, when disabled makes repair last 10 seconds instead of 60
  - Misc / Flight / Enable anti-gravs if artron depleted (false by default)
  - Admin / Artron Energy / 
    - Enable artron energy
    - Max. artron energy
    - Artron energy level at spawn (0, 25, 50, 75 or 100%) - 25% by default

- Stopped cloak from working when power is disabled (#651)
- Bug fix: flight doesn't enable in the TARDIS after interrupting teleport 

- Closes #654


lua/entities/gmod_tardis/modules/sh_artronenergy.lua
```diff
@@ -0,0 +1,311 @@
+if SERVER then
+
+    local artron_values = {
```
AmyJeanes: I think this would be good to store in a TARDIS.SomethingTable so that other modders / advanced users can choose to override values from this if they wish as it would be impossible right now.

lua/entities/gmod_tardis/modules/sh_artronenergy.lua
```diff
@@ -0,0 +1,311 @@
...
+            self:InterruptTeleport()
+            return
+        end
+        if self:GetPower() then
+            if self:GetData("flight") and TARDIS:GetSetting("flight_to_float_if_no_artron", self:GetCreator()) then
+                self:SetData("floatfirst", true)
+            end
+            self:SetPower(false)
+            for k,_ in pairs(self.occupants) do
+                TARDIS:ErrorMessage(k, "Artron Depleted.")
```
AmyJeanes: This should be using language system, suggested English text: "Artron energy depleted"

lua/entities/gmod_tardis/modules/sh_artronenergy.lua
```diff
@@ -0,0 +1,311 @@
...
+        if not fast and artron < -artron_values.cost_demat then
+            return true
+        end
+    end
+
+    ENT:AddHook("ShouldFailDemat", "artron", function(self, force)
+        if not TARDIS:GetSetting("artron_energy") then return end
+        if self:GetData("hads-attempt") then return end
+
+        if ArtronDematCheck(self) == true then
```
AmyJeanes: No need to check == true here
```suggestion
        if ArtronDematCheck(self) then
```

lua/entities/gmod_tardis/modules/sh_artronenergy.lua
```diff
@@ -0,0 +1,311 @@
...
+        local float = self:GetData("floatfirst")
+        local warning = self:GetData("health-warning")
+
+        local change = 0
+
+        if teleport then
+            return
+        elseif vortex then
+            change = change + artron_values.spend_vortex_teleport
+        elseif flight then
```
AmyJeanes, line 152: I think there should be a small energy use for having flight on but increase this when we are actively holding the movement controls and same for when boosting it should only drain extra when moving also

lua/entities/gmod_tardis/modules/sh_artronenergy.lua
```diff
@@ -0,0 +1,311 @@
...
+    end)
+
+
+
+
+    --
+    -- Conditions for artron changing
+    --
+
+    ENT:AddHook("CanIncreaseArtron", "interrupt-cooldown", function(self)
```
AmyJeanes: This should be moved into the appropriate teleport module

lua/entities/gmod_tardis/modules/sh_artronenergy.lua
```diff
@@ -0,0 +1,311 @@
...
+        if self:GetData("teleport-interrupted") then
+            return false
+        end
+    end)
+
+    ENT:AddHook("ShouldUpdateArtron", "repair", function(self)
```
AmyJeanes: This should be moved to repair module

lua/entities/gmod_tardis/modules/sh_artronenergy.lua
```diff
@@ -0,0 +1,311 @@
...
+        if self:GetData("repair-primed") or self:GetData("repairing") then
+            return false
+        end
+    end)
+
+    ENT:AddHook("ShouldUpdateArtron", "redecoration", function(self)
```
AmyJeanes: This should be moved to redecoration / repair module

lua/entities/gmod_tardis/modules/sh_artronenergy.lua
```diff
@@ -0,0 +1,311 @@
...
+        if self:GetData("redecorate")
+            or self:GetData("redecorate_parent")
+            or self:GetData("redecorate_child")
+            or self:GetData("is_redecorate_child")
+        then
+            return false
+        end
+    end)
+
+    ENT:AddHook("ShouldUpdateArtron", "hads", function(self)
```
AmyJeanes: This should be moved to hads module or closest to it

lua/entities/gmod_tardis/modules/sh_artronenergy.lua
```diff
@@ -0,0 +1,311 @@
...
+
+
+    --
+    -- Emergency artron behaviour
+    --
+    function ENT:ForceAddArtron()
+        self:Explode(30)
+        self.interior:Explode(30)
+        self:AddArtron(artron_values.increase_engine_release)
+        local newhealth = self:GetHealth() * math.random(60, 85) * 0.01
```
AmyJeanes: This results in you being able to spam to get free energy when the TARDIS is at very low health and doesn't seem right - I don't think a percentage based approach is correct here probably just a random amount of health instead

lua/entities/gmod_tardis/modules/sh_artronenergy.lua
```diff
@@ -0,0 +1,311 @@
...
+        self:ChangeHealth(newhealth)
+    end
+
+    function ENT:EngineReleaseVortexArtron(callback)
```
AmyJeanes: Callback functions are not required here as it can return instantly, refactor this to just return `true` or `false`

lua/entities/gmod_tardis/modules/sh_artronenergy.lua
```diff
@@ -0,0 +1,311 @@
...
+        if artron_values.cost_controls[control] then
+            self:AddArtron(artron_values.cost_controls[control])
+        end
+    end)
+
+    ENT:AddHook("HandleNoDemat", "artron", function(self)
+        if not TARDIS:GetSetting("artron_energy") then return end
+
+        if ArtronDematCheck(self) == true then return end
+        self:AddArtron(artron_values.cost_failed_demat)
```
AmyJeanes: This happens even if the power is off so you can spam click to teleport and lose all your energy

lua/entities/gmod_tardis/modules/sh_artronenergy.lua
```diff
@@ -0,0 +1,311 @@
...
+        if self:GetData("artron-val",0) > val then
+            self:SetArtron(val)
+        end
+    end)
+
+
+    --
+    -- Emergency artron behaviour
+    --
+    function ENT:ForceAddArtron()
```
AmyJeanes, line 284: This works even when health is disabled, what should we do in that scenario? The notification implies health has been decreased to compensate even if it hasn't. Additionally it is possible to put the TARDIS into a health warning mode even with health disabled using this control,

Calls to SetHealth should do nothing when it's turned off and turning it off should max health out like we do for adjusting the artron max energy

lua/entities/gmod_tardis/modules/sh_artronenergy.lua
```diff
@@ -0,0 +1,311 @@
...
+
+        if self:GetData("teleport") or self:GetData("vortex") then
+            self:InterruptTeleport()
+            return
+        end
+        if self:GetPower() then
+            if self:GetData("flight") and TARDIS:GetSetting("flight_to_float_if_no_artron", self:GetCreator()) then
+                self:SetData("floatfirst", true)
+            end
+            self:SetPower(false)
```
AmyJeanes, line 72: Now that power can be disabled automatically and then re-enabled without the user having to repair or exit flight mode I think we should have a new binding maybe under `P` by default to toggle power as it is frustrating in flight mode to have to exit, turn power back on and then go back into flight mode, or bring up the virtual console

lua/entities/gmod_tardis/modules/sh_artronenergy.lua
```diff
@@ -0,0 +1,311 @@
+if SERVER then
+
+    local artron_values = {
+        cost_demat = -540,
+        cost_mat = -360,
+        cost_full = -1200,
+        cost_controls = {
+            ["cloak"] = -720,
```
AmyJeanes: Should these controls use energy to turn them off as well as turn them on?

lua/tardis/cl_hud.lua
```diff
@@ -2,41 +2,73 @@
...
-    if not TARDIS:GetSetting("health-enabled") then return end
-    if not (LocalPlayer():GetTardisData("interior") or LocalPlayer():GetTardisData("exterior")) then return end
-    local tardis = LocalPlayer():GetTardisData("exterior")
-    if not IsValid(tardis) then return end
-    local sw = ScrW()
-    local sh = ScrH()
-    local health = math.ceil(tardis:GetHealthPercent())
+local function CreatePercentageHUDPanel(text, value, offset, red_level)
+
+    red_level = red_level or 20
```
AmyJeanes: Given that this is only called by two things and cannot be called by anything else as it's local we can just pass this in from the health one too, no need for this check

lua/tardis/cl_hud.lua
```diff
@@ -2,41 +2,73 @@
...
-    local y = (ScrH()-height)*0.025
+    local height = (ScrW() >= 800) and 95 or 85
+
+    if value >= 10 then width = width + 10 end
+    if value == 100 then width = width + 20 end
+
+    local x = (ScrW() - width) * 0.02 + offset
+    local y = (ScrH() - height) * 0.025
+
+    local header_font = "TARDIS-HUD-Small"
```
AmyJeanes: We can do some optimisations in this area like bringing out some of the static strings e.g. this one `header_font` as well as the others e.g. `TARDIS-HUD-Large/Med` which will reduce memory allocations, and calls to `NamedColor`

We can simply define them above as local variables outside the function scope and reference them within. I wouldn't normally bother with such micro optimisations but given this runs every frame potentially multiple times with artron and health enabled it's good to do optimise where possible

lua/tardis/cl_hud.lua
```diff
@@ -2,41 +2,73 @@
...
-    draw.DrawText( tostring(health) .. "%", healthfont, x+10, y+45, textcolor, TEXT_ALIGN_LEFT )
+    draw.DrawText( text, header_font, x+10, y+10, textcolor, TEXT_ALIGN_LEFT )
+    draw.DrawText( tostring(value) .. "%", value_font, x+10, y+35, textcolor, TEXT_ALIGN_LEFT )
+
+    return width, height
+end
+
+hook.Add("HUDPaint", "TARDIS-HUD", function()
+    local ply = LocalPlayer()
+    if not (ply:GetTardisData("interior") or ply:GetTardisData("exterior")) then return end
```
AmyJeanes: We can set the `tardis` variable up here to save us grabbing it twice and also we don't need to check the interior one at all, as that will break usage on no-interior scenarios

lua/tardis/languages/en.lua
```diff
@@ -93,6 +93,7 @@ T.Phrases = {
     ["Common.Save"] = "Save",
     ["Common.Set"] = "Set",
     ["Common.TARDIS"] = "TARDIS",
+    ["Common.ARTRON"] = "ARTRON",
```
AmyJeanes: This should be `Common.Atron`, `TARDIS` is only uppercase above as it is an acronym

lua/tardis/languages/en.lua
```diff
@@ -122,6 +123,7 @@ T.Phrases = {
     ["Controls.EngineRelease"] = "Engine Release",
     ["Controls.EngineRelease.FailedDemat"] = "Failed to dematerialise",
     ["Controls.EngineRelease.ForceDemat"] = "Force dematerialisation triggered",
+    ["Controls.EngineRelease.ArtronAdded"] = "Health has been wasted to charge artron energy in the vortex",
```
AmyJeanes: I think it would be better to say 'sacrificed' rather than 'wasted', also need to consider what this says when health is not enabled as in the other comment

lua/tardis/languages/en.lua
```diff
@@ -359,6 +361,15 @@ T.Phrases = {
     ["Settings.Sections.Global.EnableHealth.Description"] = "Should the TARDIS have health and take damage?",
     ["Settings.Sections.Global.MaxHealth"] = "Max Health",
     ["Settings.Sections.Global.MaxHealth.Description"] = "Maximum ammount of health the TARDIS has",
+    ["Settings.Sections.Global.WaitLongForRepair"] = "Enable long repair",
+    ["Settings.Sections.Global.WaitLongForRepair.Description"] = "Should the player have to wait 60 seconds for repair to finish?",
```
AmyJeanes: This language is unclear as the repair takes 'up to' 60 seconds depending on the amount of health being repaired and max health. I think it would be better to instead change it to Quick Repair and invert the setting and simply describe it as quicker repair times

lua/tardis/settings/sh_global_settings.lua
```diff
@@ -44,6 +44,93 @@ TARDIS:AddSetting({
...
+        flags = CONVAR_FLAGS,
+        desc = "TARDIS - Max amount of artron energy",
+    },
+
+    option = true,
+    section=SETTING_SECTION,
+    subsection="ArtronEnergy",
+    name="ArtronEnergyMax",
+})
+
```
AmyJeanes, line 133: Should we have an option for charging speed multipler? I think combined with these other two options that will give players/admins enough control over energy to do what they like, max health already effectively controls the decrease rate.

lua/tardis/settings/sh_global_settings.lua
```diff
@@ -44,6 +44,93 @@ TARDIS:AddSetting({
...
+        desc = "TARDIS - Initial amount of artron energy",
+    },
+
+    option=true,
+    section=SETTING_SECTION,
+    subsection="ArtronEnergy",
+    name="ArtronEnergyStart",
+})
+
+TARDIS:AddSetting({
```
AmyJeanes, line 109: Should we instead consider making this a slider for drain rate multiplier instead with the default at 1? I think that might be easier for users to understand than a random number

lua/tardis/controls/sh_control_engine_release.lua
```diff
@@ -10,6 +10,11 @@ TARDIS:AddControl({
                 TARDIS:ErrorMessage(ply, "Controls.EngineRelease.FailedDemat")
             end
         end)
+        self:EngineReleaseVortexArtron(function(result)
```
AmyJeanes: You can spam this control and gain tons of artron energy quickly, should this be possible to do?

lua/entities/gmod_tardis/modules/sh_artronenergy.lua
```diff
@@ -0,0 +1,311 @@
...
+
+        -- every 1 second:
+        spend_vortex_teleport = -32,
+        spend_flight = -45,
+        spend_flight_boost = -45,
+        spend_cloak = -36,
+        spend_cloak_handbrake = -12,
+
+        -- every 5 seconds:
+        increase_handbrake = 180 * 5,
```
AmyJeanes: This function is fine but it seems almost necessary as otherwise charging is quite slow and we do not tell the users about this at all - should we consider some kind of hint when charging and energy is low maybe? This links in with #176

lua/entities/gmod_tardis/modules/sh_artronenergy.lua
AmyJeanes, line 152: A note on flight in general: I thought we were going to disable flight if the handbrake was used? I might have missed a conversation on this though

AmyJeanes (changes requested):
This overall looks really good and an excellent addition to the TARDIS I'm sure many will have been waiting for! It appears to function in-game well for the most part although there is some unusual behaviour in a couple scenarios I have put comments in for.

As for balancing, it feels reasonable to me but I'd really like to see others give it a go as I don't really 'play with' the TARDIS as others do as I only really use it to develop on it.

lua/tardis/controls/sh_control_engine_release.lua
parar020100: you'll kill your TARDIS rather soon like that
but I might add a small delay

lua/entities/gmod_tardis/modules/sh_artronenergy.lua
parar020100: I've been thinking about that, but we'd need a whole separate thing for that
(I don't think hints would do the trick)

lua/entities/gmod_tardis/modules/sh_artronenergy.lua
parar020100: I kinda forgot that
Will do

lua/entities/gmod_tardis/modules/sh_artronenergy.lua
parar020100: Done

lua/entities/gmod_tardis/modules/sh_artronenergy.lua
parar020100: Fixed

lua/entities/gmod_tardis/modules/sh_artronenergy.lua
parar020100: Fixed

lua/tardis/languages/en.lua
parar020100: Fixed

lua/tardis/settings/sh_global_settings.lua
parar020100: I've scaled down the numbers
The default value is now 100, the min. and max. values are 25 and 400
I didn't want to add %, since it would create confusion with "start value" persentages
But it should be more intuitive

lua/entities/gmod_tardis/modules/sh_artronenergy.lua
parar020100: Done

lua/entities/gmod_tardis/modules/sh_artronenergy.lua
parar020100: Done

lua/entities/gmod_tardis/modules/sh_artronenergy.lua
parar020100: Done

lua/entities/gmod_tardis/modules/sh_artronenergy.lua
parar020100: Done

lua/entities/gmod_tardis/modules/sh_artronenergy.lua
parar020100: Should be done, but needs double-checking

lua/tardis/settings/sh_global_settings.lua
parar020100: I don't really like this idea
The point of artron was making people use handbrake and add some difficulty to the game
(allowing them to disable it if they want)

I think with such an option everyone will just make it recharge quickly and it will lose its whole point

lua/tardis/controls/sh_control_engine_release.lua
parar020100: I've done a lot of testing
With the current parameters and default health amount, you kill your TARDIS in 4-6 clicks of this one
So this can not be abused

lua/entities/gmod_tardis/modules/sh_artronenergy.lua
parar020100: I've checked, seems to work the right way for me

lua/entities/gmod_tardis/modules/sh_artronenergy.lua
parar020100: I've made it so that artron would recharge every second with handbrake enabled, and still once in 5 seconds when it's disabled. That would make it more obvious for anyone who uses handbrake at least once

lua/entities/gmod_tardis/modules/sh_artronenergy.lua
parar020100: Done

lua/tardis/controls/sh_control_engine_release.lua
AmyJeanes: Okay

lua/tardis/languages/en.lua
```diff
@@ -93,6 +93,7 @@ T.Phrases = {
     ["Common.Save"] = "Save",
```
AmyJeanes: Language files need to be updated to match these changes or they'll get wiped out

lua/entities/gmod_tardis/modules/sh_artronenergy.lua
parar020100: I believe 1bda7118e747033950399461db00285ded438def solved this

lua/entities/gmod_tardis/modules/sh_artronenergy.lua
parar020100: Fixed: 44a3043c44b8123d1c5d8e1ca754a230173f1bf7

lua/tardis/languages/en.lua
parar020100: e27baa3093c9153aef40eefea879a608d98ae0bd

lua/tardis/languages/en.lua
parar020100: This was fixed in e748a8cf338fed20c33127030d0791464cda6d5c

lua/tardis/cl_hud.lua
parar020100: befcee8d17bc313a91cad4e7940cbcbff17c1015

lua/tardis/cl_hud.lua
parar020100: befcee8d17bc313a91cad4e7940cbcbff17c1015

lua/tardis/cl_hud.lua
parar020100: befcee8d17bc313a91cad4e7940cbcbff17c1015

lua/entities/gmod_tardis/modules/sh_artronenergy.lua
parar020100: Not true
Using the throttle doesn't waste the energy since the control is not activated

lua/entities/gmod_tardis/modules/sh_artronenergy.lua
parar020100: It was intended to be like this, so that you can (for instance) be stuck in float mode
I'm unsure about this, but I'd rather give it a go, and only change if it gets lots of negative feedback

lua/entities/gmod_tardis/modules/sh_artronenergy.lua
parar020100: b390b71908ed317de4c4c2685f64bc68ffec0cf1
I've done some safeguards for that both for Engine Release and for changing health in general

lua/entities/gmod_tardis/modules/sh_artronenergy.lua
parar020100: 96c8327d45ad4246d3f38e71664b525e71192c8d
done

lua/entities/gmod_tardis/modules/sh_artronenergy.lua
AmyJeanes: As discussed not done

lua/entities/gmod_tardis/modules/sh_artronenergy.lua
AmyJeanes: In a TARDIS with lower default health this can kill it in one hit, should this be scaled by the max TARDIS health in some way maybe?

lua/entities/gmod_tardis/modules/sh_artronenergy.lua
AmyJeanes: A more general point I suppose is should enabling artron energy even be possible if health is disabled, is that a scenario we want to support?

lua/entities/gmod_tardis/modules/sh_artronenergy.lua
AmyJeanes: Okay

lua/tardis/cl_hud.lua
```diff
@@ -2,41 +2,74 @@
...
+    if ( GetConVarNumber( "cl_drawhud" ) == 0 ) then return end
+    local tardis = ply:GetTardisData("exterior")
+    if not tardis then return end
+
+    local draw_health = TARDIS:GetSetting("health-enabled")
+    local draw_artron = TARDIS:GetSetting("artron_energy")
+
+    if not draw_health and not draw_artron then return end
+
     if not IsValid(tardis) then return end
```
AmyJeanes: We are checking the TARDIS above on line 46 if it exists and then checking if it IsValid down here, we should move this up and replace the check on line 46

lua/tardis/settings/sh_global_settings.lua
AmyJeanes: I think this works for now

lua/tardis/settings/sh_global_settings.lua
AmyJeanes: I'll discuss this with you further on discord I think it is a good idea still

lua/entities/gmod_tardis/modules/sh_artronenergy.lua
parar020100: Why not?
In all scenarios except for engine release, artron energy is independent from health
It's not like we can't use one without the other
(you just have to be more attentive about when you want to materialise in the vortex, but it still works)

lua/entities/gmod_tardis/modules/sh_artronenergy.lua
parar020100: Done.
:P
800c694ac32d8cf3fc1c8a25ff530b07c430ac68

lua/entities/gmod_tardis/modules/sh_artronenergy.lua
parar020100: Fixed in a3681a5166833bc0a00569471611e292faf0883a

lua/tardis/cl_hud.lua
parar020100: Fixed 226ea43a4eafd4c6e1bab819a3bd07fb890b3a56

lua/tardis/settings/sh_global_settings.lua
parar020100: I've been giving it some thought and decided to make it in a different form
See 30340f729ea6f23cac310a1f3a44914406d5d014 and 8db80d0c985561170913e601499f62fd038f8218

It regulates the charging speed but doesn't let the user abuse it too much, allowing to select between 2 and 10 seconds for normal non-handbrake charging period (5 is default)

Handbrake charges it every second and does it rather quickly, I don't think we should modify the speed with it

lua/tardis/languages/en.lua
parar020100: Should be done

lua/entities/gmod_tardis/modules/sh_artronenergy.lua
AmyJeanes: That's fine, we just need to make sure as part of testing that it works correctly when health is disabled

AmyJeanes (approved):

PR #723, opened 2022-11-23
A tool for debugging textures

parar020100:
Just a small tool I've made for developing extensions
Outputs the list of textures / materials in convenient form with a specific convar set to 1


AmyJeanes (approved):
Looks good!

PR #755, opened 2022-12-16
Update Simplified Chinese translations

Fox2shadow:


AmyJeanes (approved):
Hey thanks for the updated translations - in future though I'd recommend updating them through CrowdIn so they go through the proper system: https://crowdin.com/project/tardis-gmod

PR #763, opened 2022-12-24
Minor fixes from the artron branch

parar020100:
Some of the fixes were presumed to be included in the update
Since we've decided to delay the artron release, I'd like them to be included


lua/entities/gmod_tardis/modules/sh_health.lua
```diff
@@ -112,7 +112,8 @@ if SERVER then
         self:SetLocked(true,nil,true)
         local maxhealth = TARDIS:GetSetting("health-max")
         local curhealth = self:GetData("health-val")
-        local time = CurTime() + ( math.Clamp((maxhealth - curhealth) * 0.1, 1, 60) )
+        local maxtime = TARDIS:GetSetting("long_repair") and 60 or 10
+        local time = CurTime() + ( math.Clamp((maxhealth - curhealth) * 0.1, 1, maxtime) )
```
AmyJeanes: The 0.1 should also probably be scaled appropriately, maybe 0.015 when long repair is disabled

lua/tardis/settings/sh_global_settings.lua
```diff
@@ -44,6 +44,22 @@ TARDIS:AddSetting({
...
+TARDIS:AddSetting({
+    id="long_repair",
+    type="bool",
+    value=true,
+
+    class="global",
+    convar = {
+        name = "tardis2_long_repair",
+        flags = CONVAR_FLAGS,
+        desc = "TARDIS - Whether repairing process lasts 60 seconds",
```
AmyJeanes: This should clarify that it is 'up to' 60 seconds, and also that having it off will make it 'up to' 10 seconds

Additionally, this should be using the language features

lua/tardis/settings/sh_global_settings.lua
```diff
@@ -44,6 +44,22 @@ TARDIS:AddSetting({
...
+    },
+
+    option = true,
+    section=SETTING_SECTION,
+    name="WaitLongForRepair",
```
AmyJeanes, line 61: This setting does not exist in the language files

AmyJeanes (changes requested):
Most fixes look good but the new long repair option has a couple of issues

AmyJeanes (approved):
All looking good, nice work!

PR #782, opened 2023-01-13
A portals debug tool + multiple bug / error fixes

parar020100:
- The portal debug tool
  - use `tardis2_debug_portals` command to enable (while looking at a portal you want to edit)
  - allows to edit portals in the game (you don't even need to resoawn the TARDIS)
  - has an option that prints the code for the portal into the console
  
- Made the context menu version of the TARDIS UI window
  - Can hide automatically when the context menu is hidden (when you stop pressing "C")
  - Has a close button at the top and a "popup" button in the interface (that allows to make it open constantly)
  - Allows less clicks for opening & using it
  - Using the other ways to open the UI (like middle mouse click in third person) open the normal version of the UI that doesn't close
  
- Tweaked the popup screen size proportions (these are more convenient since they block a smaller part of the screen horizontally)
- Replaced the "big popup UI" setting with "popup UI scale"
- Fixed the doors transparency being different from the exterior
- Temporary fix for the missing keybind error (TODO: add a prompt / warning ?)


i18n/languages/en.json
```diff
@@ -358,8 +358,8 @@
     "Settings.Sections.Global.WaitLongForRepair.Description": "Should the max repair time be 60 seconds instead of 15?",
     "Settings.Sections.Global.WaitLongForRepair": "Enable long repair",
     "Settings.Sections.Global": "Admin Options",
-    "Settings.Sections.GUI.BigPopup.Description": "Should the popup for new visual GUI be bigger?",
-    "Settings.Sections.GUI.BigPopup": "GUI big popup",
+    "Settings.Sections.GUI.PopupScale.Description": "Allows to scale the popup for new visual GUI",
```
AmyJeanes: Can you ensure that these language strings are alphabetical within the file?

lua/tardis/sh_screen_ui.lua
```diff
@@ -244,28 +244,37 @@ function TARDIS:RemoveHUDScreen()
...
+        frame:MakePopup()
+    end
+
     self.screenpopframe=frame
+    self.screen_in_context_menu = (window ~= nil)
 
     local screen = vgui.Create("DPanel",frame)
     screen.id="pop"
-    screen.width=825
+    screen.width=625
```
AmyJeanes: Curious what the reason was for changing the width here? I thought I noticed it seems a little less wide in-game

lua/tardis/libraries/libraries/sh_debug_portals.lua
```diff
@@ -0,0 +1,316 @@
...
+            portal:SetInverted(net.ReadBool())
+        end
+    end)
+else
+    function TARDIS:ShowPortalDebugMenu(p)
+        if IsValid(p.debug_window) then
+            if not g_ContextMenu:IsVisible() then
+                g_ContextMenu:Open()
+            end
+            --p.debug_window:Center()
```
AmyJeanes: Remove commented out code

lua/tardis/libraries/libraries/sh_debug_portals.lua
```diff
@@ -0,0 +1,316 @@
...
+            eao_y = val
+            UpdatePortalExitOffset()
+        end)
+        local eaor, eaor2 = SetupProperty( "Exit Point", "Roll", eao_r, 360, function(val)
+            eao_r = val
+            UpdatePortalExitOffset()
+        end)
+
+
+        local ep_category = pr:GetCategory("Exit Point")
```
AmyJeanes: I think clarifying this is an offset for the exit point a bit better might be useful, I was a bit confused as to what this did for a while

lua/tardis/libraries/libraries/sh_debug_portals.lua
```diff
@@ -0,0 +1,316 @@
...
+        TARDIS:ShowPortalDebugMenu(p)
+    end)
+end
+
+concommand.Add("tardis2_debug_portals", function(ply,cmd,args)
+    if not ply:IsAdmin() then return end
+
+    local portal = wp.GetFirstPortalHit(ply:EyePos(), ply:EyeAngles():Forward())
+
+    if IsValid(portal.Entity) then
```
AmyJeanes, line 391: Is it worth printing a message to the user if they are not pointing at a portal?

lua/tardis/libraries/libraries/sh_debug_portals.lua
```diff
@@ -0,0 +1,316 @@
...
+                print("\tinverted = " .. tostring(inverted) .. ",")
+            end
+
+            print("),")
+        end
+    end
+
+    net.Receive("TARDIS-Debug-Portals", function()
+        local p = net.ReadEntity()
+        TARDIS:ShowPortalDebugMenu(p)
```
AmyJeanes, line 382: Might be good to ensure p is valid here just in case

AmyJeanes (changes requested):
Wow nice work! The portal debug tool is really something amazing and will make it much more useful! I can see something like this for part positioning also being useful too 😉 

I've got a couple of questions and some minor things to go through before merging but overall looks really great 😄

lua/tardis/sh_screen_ui.lua
parar020100: Was:
image

Now:
image

Since it can be used in the context menu, we can keep it open even while flying
And it's better if it doesn't block the view by default

lua/tardis/libraries/libraries/sh_debug_portals.lua
parar020100: Done

lua/tardis/libraries/libraries/sh_debug_portals.lua
parar020100: Done

lua/tardis/libraries/libraries/sh_debug_portals.lua
parar020100: Done

lua/tardis/libraries/libraries/sh_debug_portals.lua
parar020100: Should be fixed

lua/tardis/sh_screen_ui.lua
parar020100: I made it a bit wider eventually, but not as wide as it used to be

i18n/languages/en.json
parar020100: Done

AmyJeanes (approved):
Looking good!

PR #784, opened 2023-01-20
Message overhaul: msg_queue + messages using von instead of functions

### LINKED ISSUE #334: Clients with high latency get metadata indexed before it exists
```
[ERROR] lua/entities/gmod_tardis/modules/sh_lock.lua:71: attempt to index field 'metadata' (a nil value)
1. unknown - lua/entities/gmod_tardis/modules/sh_lock.lua:71
2. func - lua/entities/gmod_tardis/modules/libraries/sh_message.lua:34
3. unknown - lua/includes/extensions/net.lua:32 
```

Adding a check for the variable being existing before anything is done should be enough

Known error entry points:
- sh_lock.lua:71 (85)
- sh_power.lua:32
- sh_doors.lua:230

### LINKED ISSUE #778: Networking race conditions: clientside entities not fully initialized when the server calls them
As noticed in multiple cases, sometimes the server calls `SendMessage()` before the client entity is fully initialized 

From discussion with Matt:
> nah what you'll want to do is handle the net.Receive on the clientside that powers SendMessage
> and basically queue up messages if the exterior doesnt exist yet
> we do something like this for part initialisation already

> in modules/libraries/sh_message.lua

```lua
net.Receive("TARDIS-MessageExt", function(len,ply)
    local ent=net.ReadEntity()
    local name=net.ReadString()
    if IsValid(ent) and messagehandlers[name] then
        messagehandlers[name](ent,ply)
    end
end)
```
> you should hold if the special property _init is not set on the entity

> it's a similar idea for clientside initialisation of parts
```lua
if parent._init then
    TARDIS:SetupPart(e,name,ext,int,parent)
else
    if not parent.partqueue then parent.partqueue = {} end
    parent.partqueue[e] = name
end
```
> the reason is that the order of entities that are sent via source networking is not guaranteed in any way when entities are created
> so the client may be aware of the door before it is aware of the exterior
> it's part of the reason why joining a game with an existing tardis causes issues too
> it can be fixed with stuff like queues

Another suggested thing: perhaps the PostInitialize hook should only run after both server and client have fully initialized the entities?

parar020100:


AmyJeanes (approved):
All working nicely, great job!!

PR #788, opened 2023-01-22
Teleport module refactoring

### LINKED ISSUE #781: Handbrake during the beginning of remat still results in mat at the destination point
Discovered by @Artixc and @CrethFox (iirc)

> While starting to remat if handbreak is pulled, tardis will normal fall out of sky and remat at its destination without animation and glitch

> Must pull handbreak before tardis exits vortex

https://user-images.githubusercontent.com/32988384/211487536-967faab9-727f-4d87-9d03-d394d4fd78aa.mp4

parar020100:
This also fixes #781


lua/entities/gmod_tardis/modules/cl_effects.lua
```diff
@@ -0,0 +1,46 @@
...
+		self.effect_pos = self.interior:GetPos() + Vector(0, 0, 50)
+	end
+end
+
+function ENT:InteriorExplosion()
+	if self.effect_pos == nil then
+		get_effect_pos(self)
+	end
+
+	local function rand_offset() return math.random(-40, 40) end
```
AmyJeanes: `rand_offset` function appears twice with slightly different values, this should be consolidated into the one function. If you need to pass different parameters e.g. 35 vs 40 make it a parameter to the function

lua/entities/gmod_tardis/modules/teleport/sh_tp_attached.lua
```diff
@@ -0,0 +1,106 @@
...
+        self:SetData("demat-attached",nil,true)
+    end)
+
+    ENT:AddHook("PreTeleportPositionChange", "attached_hoverballs", function(self, pos, ang, phys_enable)
+        local attached=self:GetData("demat-attached")
+        if attached then
+            for k,v in pairs(attached) do
+                if IsValid(k) and not IsValid(k:GetParent()) then
+                    k.telepos=k:GetPos()-self:GetPos()
+                    if k:GetClass()=="gmod_hoverball" then -- fixes hoverballs spazzing out
```
AmyJeanes, line 53: Been a long time since I wrote this comment 😅 the wording should change here to be inoffensive now. Maybe 'fixes hoverballs freaking out'?

lua/entities/gmod_tardis/modules/teleport/sh_tp_failed.lua
```diff
@@ -0,0 +1,215 @@
+-- Failed teleport, interrupted teleport, effects
```
AmyJeanes: Comment may not be accurate now as interrupted teleport is now in a different file?

AmyJeanes (changes requested):
Mostly looking good to me! I just wanted to also confirm that the only actual changes to behaviour were in the other 2 commits and not part of "Moved files" or "Refactoring"?

lua/entities/gmod_tardis/modules/teleport/sh_tp_failed.lua
AmyJeanes: Also, we should also have little comments at the top of each file in here, some appear to be missing them

lua/entities/gmod_tardis/modules/teleport/sh_tp_attached.lua
AmyJeanes: After discussion this is fine actually, the term "spazzing out" refers to things like muscle spasms and I think that is clear enough given the context here

AmyJeanes (approved):
Nice work!

PR #793, opened 2023-01-23
HUD overhaul

parar020100:
I've tried many options and I'm pretty sure this is the best one

+ I've also decided that it's better to show artron energy by value, not by percent
(since max value is between 50 and 400)

Features animated pulsing if the value is red

image

image

image

image
image
image
image

image
image


parar020100:
I've also experimented with small font options, but none of them turned good enough
And it's not a thing worth spending much time on

AmyJeanes (approved):

PR #798, opened 2023-01-28
Fade effect on entering with walk key(parar's suggestion)

PoogieTheGoose:
Made a small screen fade effect on entering or exiting the TARDIS with walk key(ParAr's suggestion)
Also made the fade effect on enter a bit transparent so i can see where you go the second you enter the interior.Small change but i think it's a nice addition


AmyJeanes (commented):
I've adjusted the effect to be black instead of white and removed the unused fifth parameter that is being passed in, it seems a lot better than giving you a flashbang when you enter / exit the TARDIS. Let me know if you're happy with these changes and I'll get this in.


https://user-images.githubusercontent.com/2363642/215337698-a7daa4bc-e416-409c-ac13-4fcd729dfab5.mp4

AmyJeanes (approved):
Happy for now as discussed in Discord, will revisit if we can thing of a better effect

Issue #28, opened 2015-02-17
Spacebuild (and similar addons) compatibility


MattJeanes:
If you turned engines off in zero-g on spacebuild, it would fall with gravity. Need to add checks to prevent this behaviour.

Also, could do with life support being present in the interior if possible.

MattJeanes:
[Environments](http://steamcommunity.com/sharedfiles/filedetails/?id=104484308)

iNoPeRaTiVe:
If it's possible, when life support is turned off could gravity be turned off in the TARDIS? I don't know if that's something you can do.

MattJeanes:
Turning gravity off on power-off would be a nice idea

Noname12345GER:
Give objects in the TARDIS a slight applyforce upwards to make that cozy, chaotic space feeling. Ceiling collision is a Problem then.

theangelcastiel:
Should also be a button that makes an oxygen shell around the TARDIS. Small though, so it's not OP

theangelcastiel:
Possibly the instant the TARDIS dematerializes the set destination should have said oxygen shell too

parar020100:
![image](https://user-images.githubusercontent.com/32988384/213815908-9c3c3e52-dadf-4d50-aaa8-d607bc9101d0.png)
![image](https://user-images.githubusercontent.com/32988384/213815917-a74764b5-375d-4076-9a2b-14a31cb3249f.png)

PR #805, opened 2023-02-10
Rudimentary Life Support

jeppevinkel:
This is a small update that will prevent players from suffucating in the TARDIS when Spacebuild is enabled.

It won't create artificial gravity or an atmosphere, but it can serve as a temporary solution until #28 can be properly implemented.

### DIFF
--- a/lua/entities/gmod_tardis_interior/init.lua
+++ b/lua/entities/gmod_tardis_interior/init.lua
@@ -24,9 +24,46 @@ function ENT:Initialize()
         self.ExitDistance=self.metadata.Interior.ExitDistance
     end
     self.BaseClass.Initialize(self)
+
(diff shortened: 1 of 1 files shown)


AmyJeanes:
Hey, thanks for contributing! Which Spacebuild addon are you using to test this with? I believe there is other SB-like addons that people use these days but I've been out of the loop for a while

jeppevinkel:
Hey, yeah I've only tested this with Spacebuild 3 as I don't think Spacebuild 4 was ever finished. I don't really know about any of the other SB-like addons, so I haven't tested any of those.

AmyJeanes:
I've ended up adding my own version of spacebuild support you can see here: https://github.com/MattJeanes/TARDIS/commit/32dcf64a9f335a05000da115e5eea9ac6bc2b14c

As such I will be closing this PR. Hopefully that has everything you're looking for, but in any case thank you for contributing

jeppevinkel:
Yeah your implementation is a lot more thorough than mine, thanks for doing it 👍

PR #806, opened 2023-02-11
Poogie artron

PoogieTheGoose:
Added an engine release function sound and fixed an error that appeared when no interior option was selected


AmyJeanes (approved):
Looks good, and nice fix for the no interior option!

PR #814, opened 2023-02-17
Artixc Exteriors

Artixc:
-4 New Exteriors!
image
Mk1, Mk2, Mk3, and Sidrat
In Respective Order

-Added English translations for new options
-Removed Vtal Hellbent Exterior


AmyJeanes (approved):
Looks good! We should do a follow up on this to get classic doors working on these new exteriors for the default but it's all good for now.

PR #821, opened 2023-02-19
Screen update

### LINKED ISSUE #437: Default visGUI theme elements for each interior
Interior makers should have the opportunity to set default elements for GUI, like background or frame styles

### LINKED ISSUE #568: Simplify coordinates GUI
The current version is extremely complicated even for me. 
I keep confusing "Save" and "New" buttons, since saving current location in the list is also "saving"
I want an easier and more intuitive menu

## TODO:

* Better use of the screen space 
  * Bigger text inputs
  
* Add the display of "current" and "destination" coordinates 
  * "Copy" and "Refresh" buttons
 
* Make the button names more intuitive and obvious (what's Add / Save / Set right now? noone knows)
  * Rename "New" into "Add"
  * Remove "Update", which is never used
    * people would just remove & add if you need to update it, or name differently

* "Save" and "Load" buttons should be removed completely
  * When the location is added to the list, it is saved to the file automatically (same as music)
  * When the location is being removed, an "Are you sure?" dialogue window should appear to minimize accidents

Here's a sketch we made with @CrethFox
изображение

### LINKED ISSUE #574: Screen GUI background for extensions
#437 was closed, because it was too complicated
However, to keep integrity with the show, we should allow custom backgrounds for interiors

### LINKED ISSUE #624: Saving coordinates from text inputs is broken
After clicking "save", you can select whether the saved coordinates should be the current position or the text input values

When you select "input values", `(0,0,0)` is saved instead

> This seems to always be 0,0 no matter what, what's the intended behaviour here?
_Originally posted by @MattJeanes in https://github.com/MattJeanes/TARDIS/pull/623#discussion_r805472970_

### LINKED ISSUE #636: Allow to make the music louder
It is 1..100 now
I think we should make it 1..300 or something

### LINKED ISSUE #637: Change music volume from the music screen

### LINKED ISSUE #730: Return coordinates UI to 3D screens
We've made it to be a popup due to scaling and some other issues
I think it should be returned back for gameplay reasons, but with a different (simplier) interface

Might be related to #568

### LINKED ISSUE #811: Script errors while switching UI interface type
Reported by Zaursus in Discord (ticket 243)

> If you create a TARDIS with the old interface and switch to the new interface in the Q menu without changing the TARDIS, script errors appear

```[TARDIS] lua/tardis/sh_screen_ui.lua:160: attempt to index field 'left_arrow' (a nil value) 1. SwitchScreen - lua/tardis/sh_screen_ui.lua:160 2. DoClick - lua/tardis/sh_screen_ui.lua:515 3. OnMouseReleased - lua/vgui/dlabel.lua:237 4. fn - l...```

### PR DESCRIPTION
**Pull request by parar020100**
image

- Coordinates screen overhaul (#568)
  - New interface showing current position and destination 
    - Allowed to save current position and current destination with separate buttons
    - Added buttons to copy the pos/destination values into editable input fields
    - More intuitive buttons layout
  - Split saving current location and information from the inputs (#624 fixed)
  - Buttons get disabled in situations in which they shouldn't be used
  - Deleted locations are saved instantly (without having to save the whole list)
  - Added a teleport progress bar
  - Forbade to save current location in the vortex
  - Added random numbers in location field while in the vortex
  - Replaced one of the background images with a more in-universe one
  - Fixed script errors while trying to delete "random" from the locations list

- Force unique new location names: if the name already exists, a number will be added
  - Example: `New location`, `New location(1)`, `New location(2)`

- Music screen upgrade
  - Added a background
  - Added a panel on the right
  - Added a popup-only music volume setting slider to the music screen (#637) 
  - Allowed louder music volume values as requested many times (#636)

- Scanner screen was made fullscreen (grey frame removed) + uses the arrows from the new UI

- 3D versions of coordinates and music screens
  - A new `ListView3D` class as a `ListView` alternative for 3D screens
  - Enhanced DTextEntry3D2D vgui type (cl_vgui.lua)
    - Added "Numeric" parameter support for 3D screens
    - Fixed OnChange() hook function

- GUI themes update
  - Added support for default theme per interior
    - Usage example: `T.Interior.UI_Theme = "default-transparent"`
  - Added "Set by interior" option for GUI theme (and made it default)
  - Added support for theme's default colour
  - Small tweaks in the looks of each theme

- Made some GUI animations and button clicks faster
- Added a setting that disables UI animations

- Multiple optimization-related changes of the UI
  - Removed some Think() hooks from screen elements
  - A universal way of defining the left and right arrow button functions for each screen (should also fix #811)
  - Cleaned up code and removed unused functions (like TARDIS:PushScreen)

- Removed "Select manually" from coordinates screen

- Reduced the number of `\n` in debug output
- Added translation lines for new screen interface

Closes #568
Closes #637
Closes #624 
Closes #636
Closes #811
Closes #730
Closes #574
Closes #437

### CODE REVIEW

File: i18n/languages/en.json [hunk 1]
```diff
@@ -278,24 +278,23 @@
...
     "Screens.Coordinates": "Coordinates",
     "Screens.Coordinates.ConfirmRemoveLocation": "This will remove the selected location PERMANENTLY! Are you sure?",
-    "Screens.Coordinates.CurrentSetDestination": "Current Set Destination",
-    "Screens.Coordinates.GetCurrentPosition": "Get Current Position",
-    "Screens.Coordinates.Loaded": "Locations loaded",
+    "Screens.Coordinates.Copy": "Copy",
+    "Screens.Coordinates.CurrentPosition": "Current position:",
+    "Screens.Coordinates.Delete": "Delete",
+    "Screens.Coordinates.Destination": "Destination:",
+    "Screens.Coordinates.LocationsList": "LOCATIONS LIST",
```
AmyJeanes: Should this be normal case and then made uppercase where it is needed in the code to enforce consistency?

lua/entities/gmod_tardis_interior/modules/cl_vgui.lua
```diff
@@ -1,16 +1,31 @@
...
-            pnl:OnTextChanged()
-            pnl:OnEnter()
-            textpnl=nil
-        end,function() textpnl=nil end)
+        local old_text = pnl:GetText()
+        Derma_StringRequest(TARDIS:GetPhrase("Common.Interface"),
+            pnl.sub3D2D or pnl.strTooltipText or TARDIS:GetPhrase("Common.EnterTextInput"),
+            pnl:GetText(),
+            function(text)
+                if text ~= old_text and number_ok(pnl, text) then
```
AmyJeanes, line 16: Can you explain why we're checking for numbers here? As far as I'm aware this is a generic text entering field

lua/tardis/screens/cl_screen_scanner.lua
```diff
@@ -22,8 +22,8 @@ TARDIS:AddScreen("Scanner", {id="scanner",text="Screens.Scanner", menu=false, or
     screen.scannerfov=120
     
     local scanner=vgui.Create("DImage",frame)
-    scanner:SetSize(frame:GetWide()-screen.gap2,frame:GetTall()-screen.gap2)
-    scanner:SetPos(screen.gap,screen.gap)
+    scanner:SetSize(frame:GetWide(),frame:GetTall())
+    scanner:SetPos(0,0)
```
AmyJeanes: Can we just remove this if we're setting 0,0?

lua/tardis/screens/cl_screen_coordinates.lua
```diff
@@ -1,49 +1,483 @@
...
+    llist:SetPos( d,d )
+    llist:AddColumn(TARDIS:GetPhrase("Screens.Coordinates.LocationsList"))
+    llist:SetMultiSelect(false)
+
+    local panel_w = (w - 3 * d) / 2
+    local panel_h = (h - 4 * d) / 3
+    local panel_d = (screen.is3D2D and 0.01 or 0.02) * math.min(w,h)
+    local panel_left = 0.5 * w + 0.5 * d
+
+    local PositionPanel = vgui.Create( "DPanel",frame )
```
AmyJeanes: Use snake_case for this element

lua/tardis/screens/cl_screen_coordinates.lua
```diff
@@ -1,49 +1,483 @@
...
+    pos_x:SetSize(pos_elem_w,pos_elem_h)
+    pos_x:SetNumeric(true)
+    pos_x:SetEnabled(false)
+    pos_x:SetFont(font)
+    pos_x.Think = function(self)
+        if IsValid(ext) then
+            if ext:GetData("vortex") then
+                randomize_coordinate(self)
+            else
+                self:SetText(math.Round(ext:GetPos().x, round_digits))
```
AmyJeanes: Would there be a meaningful performance benefit here by doing these things in the Think hook of the parent panel so we don't need to call GetPos / GetAngles 3 times per frame for example

lua/tardis/screens/cl_screen_coordinates.lua
```diff
@@ -1,49 +1,483 @@
...
+    pos_copy:SetText(TARDIS:GetPhrase("Screens.Coordinates.Copy"))
+    pos_copy:SetFont(font)
+    pos_copy.Think = function(self)
+        if not IsValid(ext) then return end
+        EnsureEnabled(self, not ext:GetData("vortex"))
+    end
+
+
+
+    local DestinationPanel = vgui.Create( "DPanel",frame )
```
AmyJeanes: Use snake_case for this panel

lua/tardis/screens/cl_screen_coordinates.lua
```diff
@@ -1,49 +1,483 @@
...
+            end
+            return
+        end
+        dst_progress:SetVisible(true)
+
+        local tp_metadata = ext.metadata.Exterior.Teleport
+        local fast = ext:GetData("demat-fast")
+
+        if ext:GetData("demat") then
+            local steps = #tp_metadata.DematSequence - 1
```
AmyJeanes: Can we move this logic something into the teleport module itself and reference it from here? I can see this being useful in other places and it being there also reduces the chance that we forget to update it if we make some changes in that area

lua/tardis/screens/cl_screen_coordinates.lua
```diff
@@ -1,49 +1,483 @@
...
+    dst_copy:SetFont(TARDIS:GetScreenFont(screen, "Default"))
+    dst_copy.Think = function(self)
+        if not IsValid(ext) then return end
+        EnsureEnabled(self, ext:GetDestPos() ~= nil)
+    end
+    dst_copy:SetFont(font)
+
+
+
+    local InputPanel = vgui.Create( "DPanel",frame )
```
AmyJeanes: Use snake_case for panel name

lua/tardis/screens/cl_screen_coordinates.lua
```diff
@@ -1,49 +1,483 @@
...
+                dst_progress:SetFraction(0)
+            end
+            return
+        end
+        dst_progress:SetVisible(true)
+
+        local tp_metadata = ext.metadata.Exterior.Teleport
+        local fast = ext:GetData("demat-fast")
+
+        if ext:GetData("demat") then
```
AmyJeanes: While in the vortex it seems to easily progress and then jump backwards forever until you start landing. Generally this seems to be a bit unreliable if only an estimate, I wonder if we can improve this by merging the progress data into the actual demat process perhaps?

AmyJeanes (changes requested):
This looks great, nice work! Couple of minor comments and questions but nice big improvements to the screens :)

lua/tardis/screens/cl_screen_coordinates.lua
parar020100: ok

lua/tardis/screens/cl_screen_coordinates.lua
parar020100: ok

lua/tardis/screens/cl_screen_coordinates.lua
parar020100: ok

lua/entities/gmod_tardis_interior/modules/cl_vgui.lua
parar020100: This element is based on `DTextEntry`
DTextEntry has a numeric option defined `SetNumeric()` / `GetNumeric()` functions
This makes the 3D version work with that

As you can see above, if `pnl` is not numeric, it allows any text:
```lua
local function number_ok(pnl, text)
    return (not pnl:GetNumeric()) or (text == "") or (tonumber(text) ~= nil)
end
```

If the panel has `SetNumeric()` set to true, it will filter out non-numeric values, exactly like it does now.

In the coordinates screen, you can change the name to any text you like, but the coordinates are forced to be numbers, which helps us avoid many errors: we do `x:SetNumeric(true)` for x,y,z,pitch,yaw and roll, but `namebox` doesn't have that

lua/tardis/screens/cl_screen_scanner.lua
parar020100: I guess we can

lua/tardis/screens/cl_screen_coordinates.lua
parar020100: ok

lua/tardis/screens/cl_screen_coordinates.lua
parar020100: ok!

PoogieTheGoose: ok?

i18n/languages/en.json
parar020100: Actually thinking about it, it's okay if it's just lowercase by default

lua/tardis/screens/cl_screen_coordinates.lua
parar020100: I've made it smoother by making the start vortex position continue the demat process 
(no "jump" at the start)

Not sure whether that's what you meant

As for going backwards later, it was intended like that
(a reference to those never-ending download progress bars)

lua/tardis/screens/cl_screen_coordinates.lua
```diff
@@ -45,136 +40,116 @@ TARDIS:AddScreen("Destination", {id="coordinates", text="Screens.Coordinates", m
...
+        if not TARDIS:GetSetting("gui_animations") then
+            return "???"
+        end
+        return math.random(-99999999, 99999999) * 0.0001
+    end
+
+    position_panel.Think = function(self)
         if not IsValid(ext) then return end
-        EnsureEnabled(self, not ext:GetData("vortex"))
+        if ext:GetData("vortex") then
```
AmyJeanes: We are calling `ext:GetData("vortex")` in this function 3 times, cache it in a local variable and re-use

lua/entities/gmod_tardis_interior/modules/cl_vgui.lua
AmyJeanes: Ah okay, makes sense I thought pnl:GetNumeric() was getting the value of the text as a number which is what caused my confusion :)

lua/tardis/screens/cl_screen_coordinates.lua
parar020100: fixed

AmyJeanes (approved):

PR #830, opened 2023-02-23
Scanner for parts support

parar020100:
Quick testing: add this to Scanners{}

```lua
,
        {
            part = "default_screen",
            mat = "models/drmatt/tardis/tardismonitor",
            width = 1024,
            height = 1024,
            ang = Angle(0,0,0),
            fov = 90,
        }
```


AmyJeanes (approved):
I've made a slight tweak to make it all serverside as the part entity may not yet exist clientside and I don't want any other code trying to use it clientside unaware of this issue

PR #832, opened 2023-02-25
Multiple small fixes

### LINKED ISSUE #407: If a TARDIS redecorates, all props from inside fall out somewhere in the world

### LINKED ISSUE #625: Vortex Travel Bypass
If a player activates fast return and while specifically DEmatting, enters a location to travel to via coordinates
it will go to that location instead of the fast return location, thereby skipping vortex travel without disabling it.

We can't just stop usage of coordinates if fast return is pressed because if in multiplayer a user already has it open
then that doesn't work.
The Set button should be disabled for a set period of time once fast return is used.
Long enough for the demat to complete BUT no more time than it takes for the tardis to finish rematting.

### LINKED ISSUE #632: Redecoration fails when attempting to redecorate to a different variant of the same interior
Reported by MrPolaroid600 on Steam bug reports:

> Been playing around with the new settings system. It seems that the redecoration function bugs out when you attempt to have it to a different variant of the same interior (for example, changing the roof and exterior settings of the 1963). It doesn't seem to recognise that an interior has been selected for redecoration and therefor changes to a random one.
From my experimenting, it seems to be linked to trying to change settings while keeping the same base interior. For example if I want to keep the 1963 white version but change a couple of its options, it will either randomise to a completely different interior or apply the changes but change to the default 1963 instead of the white version.

### LINKED ISSUE #656: HADS doesn't leave the handbrake on
imo it should, so that the player has to look for it instead of always being able to summon the TARDIS

### LINKED ISSUE #657: No warning for non-existing tip style
cl_tips.lua:22 causes error 
A warning must be added to cl_tips:16

### LINKED ISSUE #704: Physlock doesn't re-enable correctly after redecoration
Just need to check it on init and make sure the physics is frozen

https://user-images.githubusercontent.com/2363642/187327918-5dbe06d9-c00f-4434-b121-88df573baef7.mp4

### LINKED ISSUE #708: Error: demat-pos ends up being a table instead of a vector
[TARDIS] lua/entities/gmod_tardis/modules/teleport/sh_tp_main.lua:126: bad argument #1 to 'WriteVector' (Vector expected, got table)
  1. WriteVector - [C]:-1
   2. func - lua/entities/gmod_tardis/modules/teleport/sh_tp_main.lua:126
    3. SendMessage - lua/entities/gmod_tardis/modules/libraries/sh_message.lua:12
     4. Demat - lua/entities/gmod_tardis/modules/teleport/sh_tp_main.lua:126
      5. ext_func - lua/tardis/controls/sh_control_teleport.lua:14
       6. Control - lua/tardis/sh_control.lua:44
        7. unknown - lua/tardis/sh_parts.lua:86

### LINKED ISSUE #734: Missing conditions for redecoration toggle
You can enable redecoration during take off, and have the TARDIS disappear afterwards
We need to add a "CanEnableRedecoration" hook and forbid redecoration while flying etc.

### LINKED ISSUE #752: Redecoration: remove separation of int and ext data

### LINKED ISSUE #762: Starting the music during redecoration results in script errors
Reported by @CrethFox 

```[tardis] lua/includes/extensions/client/panel.lua:580: Tried to use a NULL Panel!
  1. GetChildren - [C]:-1
   2. Clear - lua/includes/extensions/client/panel.lua:580
    3. onsuccess - addons/tardis/lua/tardis/screens/cl_screen_music.lua:123
     4. unknown - lua/includes/modules/http.lua:29
     
this error does not seem to have any effect on the music itself other than a minor delay that wouldn't otherwise occur 
nvm i figured it out
turns out it really does not like it if you, in order:
1) Enable Redecoration
2) Play any music (error occurs here)
3) leave the tardis

### LINKED ISSUE #809: Check that T.Base exists for each interior
From Discord discussion on ticket 237
> Artiom — 06/02/2023 23:27
we need some sort of a warning for a missing T.Base
this thing caused everyone many hours of trying to find the error
Divided — 06/02/2023 23:27
hah
yeah good idea
Artiom — 06/02/2023 23:27
it showed all kinds of errors from all over the code
I believed I was going crazy lol

> also we need a check that the base interior is somewhere in the inheritance tree
to ensure that all interiors have it merged in the end
if base isn't in the interior list, there can be metadata without T.Interior, and that's also annoying to debug

### LINKED ISSUE #816: TARDIS keeps working when killed during a HADS demat
(Reported by Zaursus)

This actually causes another weird behaviour - infinite HADS damage loop
How to reproduce:
- enable hads and trigger it
- kill the TARDIS before it fully demats
- wait till it becomes materialised again
- demat again and mat, press the handbrake during mat
- the TARDIS will get damage from the handbrake and HADS will be triggered again

this allows to get a working TARDIS with 0 health and 0 artron

### LINKED ISSUE #833: Toggling vortex flight allows to waste less artron

### PR DESCRIPTION
**Pull request by parar020100**
- Fixed a bunch of bugs, errors and issues
  - Fixed TARDIS keeping power with no health (#816)
  - Fixed destination select notifications using `TARDIS:Message` instead of `TARDIS:ErrorMessage`
  - Fixed fast return fail notification being sent twice
  - Made tips use a default `white_on_grey` style if the previously chosen one doesn't exist any more (#657)
  - Fixed redecoration working together with teleport and missing conditions (#734 & probably #762)
  - Small fixes of TT Capsule portals in templates
  - Fixed redecoration staying on when repair is cancelled
  - Fixed teleport sequences in base interior to be consistent with old ones
  - Other safeguards to prevent other types of script errors (includes #809, #708)
  
- Destination refactoring: moved some functions into `destination.lua`
  - Added a `CanChangeDestination` hook
  - Forbade to change destination during fast return

- Allowed fast return in vortex (starts mat back into old position)
- Made fast return cheaper and possible even with very low artron
  - Forbade to change destination during fast return, making it efficient but limited
- Forbade toggling vortex flight during demat and mat
- Allowed redecorating into the same interior (due to multiple versions existing) (fixes #632)
- Forbade toggling repair both during teleport and vortex
- Redecoration now doesn't disable physlock at all
- Made HADS keep the handbrake on and use force demat if it is

- Cleanup
  - removed separation of interior and exterior data in redecoration (#752)
  - removed unrequired warning about missing tip text
    - due to versions and `T.Base` inheritance, it is totally okay to have tips with no texts

Closes #816, closes #752, closes #657, closes #704, closes #708, closes #809, closes #656, closes #625, closes #833, closes #734, closes #762, closes #632

### CODE REVIEW

File: lua/entities/gmod_tardis/modules/sh_destination.lua [hunk 1]
```diff
@@ -110,6 +110,41 @@ TARDIS:AddKeyBind("destination-snaptofloor",{
...
 if SERVER then
+    function ENT:SetDestination(pos, ang)
+        if self:CallCommonHook("CanChangeDestination", pos, ang) == false then
+            return false
+        end
+
+        if not isvector(pos) or not isangle(ang) then
+            self:SetData("destination_pos",nil,true)
+            self:SetData("destination_ang",nil,true)
+            return false
```
AmyJeanes, line 121: Should we be returning false here? It is unclear if the change failed or if it just cleared the destination from the return atm

lua/entities/gmod_tardis/modules/sh_redecorate.lua
```diff
@@ -122,13 +151,14 @@ if SERVER then
...
-        if ext_saved_data then
-            for name,value in pairs(ext_saved_data) do
+        local saved_data = self:GetData("redecorate_parent_data")
+        if saved_data then
+            for name,value in pairs(saved_data) do
                 self:SetData(name, value, true)
             end
         end
-        self:SetData("redecorate_parent_ext_data", nil, true)
+        self:SetPhyslock(saved_data["physlock"])
```
AmyJeanes: Should this be handled here? It seems like this should maybe go into the physlock module as part of the MigrateData hook

lua/entities/gmod_tardis/modules/teleport/sv_tp_fastreturn.lua
```diff
@@ -23,3 +38,10 @@ ENT:AddHook("StopMat", "fastreturn", function(self)
     end
 end)
 
+ENT:AddHook("CanChangeDestination", "fastreturn", function(self, pos, ang)
+    if not self:GetData("fastreturn") then return end
+
+    if self:GetData("teleport") or self:GetData("vortex") then
```
AmyJeanes, line 44: This part of the hook should probably be elsewhere

lua/entities/gmod_tardis_interior/modules/sh_parts.lua
```diff
@@ -14,6 +14,17 @@ ENT:AddHook("Cordon","parts",function(self,class,ent)
     if ent.TardisPart then return false end
 end)
 
+if SERVER then
+    ENT:AddHook("OnRemove","inside_entities",function(self)
+        for i,v in ipairs(ents.FindInSphere(self:GetPos(), self.metadata.Interior.ExitDistance)) do
```
AmyJeanes: Use the same condition that the cordon system in Doors uses:

`for k,v in pairs(ents.FindInBox(self:LocalToWorld(self.mins),self:LocalToWorld(self.maxs))) do`

In-fact this really should be a change to Doors itself, maybe make the cordon system shared and do it in there

lua/tardis/libraries/libraries/sh_settings.lua
```diff
@@ -121,11 +121,25 @@ function TARDIS:GetSetting(id, src)
         ply = (src:IsPlayer() and src) or src:GetCreator()
     end
 
+    local function list_value_exists(setting_data, value)
```
AmyJeanes: I'm not entirely sure what's causing this but I got this a couple times after loading your branch when I spawned a TARDIS, although seemed to go away after a bit so couldn't debug any further:

Clientside error by the way
```
[tardis] addons/tardis/lua/tardis/libraries/libraries/sh_settings.lua:405: attempt to call method 'CallCommonHook' (a nil value)
  1. OnSettingChanged - addons/tardis/lua/tardis/libraries/libraries/sh_settings.lua:405
   2. func - addons/tardis/lua/tardis/libraries/libraries/sh_settings.lua:364
    3. unknown - lua/includes/extensions/net.lua:33
```

Not sure if you know what might have caused this but may be worth adding a check to ensure the function exists before calling it, looks like it's being called before it's finished initialising

AmyJeanes (changes requested):
Looking mostly good! A ton of little tweaks and fixes for sure 😄 

I've got a couple of little question and comments but overall looks and seems to work in-game pretty well, definitely want some proper testing on this though as you're touching a lot of core areas.

lua/entities/gmod_tardis/modules/teleport/sv_tp_fastreturn.lua
parar020100: the logic here is `fastreturn and (teleport or vortex)`
we cannot split it

lua/entities/gmod_tardis/modules/sh_destination.lua
parar020100: It returns whether destination is currently set rather than the success of the action
Seems more convenient, I'd prefer to keep it this way
Situations in which we clear the destination do not cause any notifications / output, so it's okay

AmyJeanes: Okay, mainly just wanted to check this was intentional and wasn't going to mess up any callers

lua/entities/gmod_tardis/modules/teleport/sv_tp_fastreturn.lua
AmyJeanes: Ah right, okay!

lua/entities/gmod_tardis/modules/sh_redecorate.lua
parar020100: fixed

lua/tardis/libraries/libraries/sh_settings.lua
parar020100: added

lua/entities/gmod_tardis_interior/modules/sh_parts.lua
parar020100: removed from TARDIS

parar020100: added to doors, see https://github.com/MattJeanes/Doors/pull/26

AmyJeanes (approved):

Issue #510, opened 2021-11-16
Teleport NPCs inside when landing on them


MattJeanes:
Reported as a bug initially but this is just missing functionality, but I think we should add this and should be an quick win

parar020100:
I believe #848 solved this

PR #848, opened 2023-03-11
Made npc's teleport inside when landing on them

PoogieTheGoose:
#510


AmyJeanes (approved):
Nice, works nicely! Nice touch with NPCs being relative to their outside position so you can land on multiple NPCs without issue

PR #849, opened 2023-03-11
Metadata / interior system refactored + legacy mode + small tweaks

### LINKED ISSUE #556: Full legacy mode
People still miss the legacies and think that some things used to be better in them. Others don't like the black portals while computers aren't strong enough to render the normal ones.
Some just like the old mechanics.

We should add a legacy mode, in which "e" just teleports into the TARDIS.

### LINKED ISSUE #666: Errors when an extension interior loads before the "base.lua" file
load order issue, for the core addon the extensions are loading before the base addon itself ( Core Addon) causing script errors to classic doors.

### LINKED ISSUE #842: Wrong warning for failed demat with doors auto-closing
Found by Za'ursus:
> If you try to enter the vortex on the TARDIS with the door open and the handbrake active, a warning appears about the impossibility of dematerializing with the door open, even if there is an auto-closing setting of the doors, instead of a warning about the time rotor handbrake turned on

parar020100:
**Features:**
- Legacy door mode (#556)

**Fixes:**
- Fixed some extensions' metadata not merging properly due to random load order (#666)
- Fixed wrong warning for failed demat with doors auto-close (#842)
- Reduced lag while reloading complicated interiors

**Internal changes:**
- Refactoring of the interior metadata system
  - interior metadata is only merged when a TARDIS is spawned now
  - merged metadata is cached, cache is cleared when the interior file updates
  - consequently, no lag when editing interior file
  - templates are added during spawn and not cached
  - spawnmenu elements (versions and custom settings) are handled separately from MetadataRaw

- sh_parts.lua had `- TODO: Improve` in it, I did my best

- small change to prevent it all breaking when LightOverride=false is specified in the extension

Closes #666
Closes #842
Closes #556


lua/tardis/interiors/modules/sh_templates.lua
```diff
@@ -88,17 +88,28 @@ function TARDIS:AddInteriorPartsRotation(template, rotate_ang)
...
-    if not cur_metadata.Templates then
-        return cur_metadata
+function TARDIS:SetupTemplateUpdates(id)
+    local t = self.MetadataRaw[id]
+    if not t.Templates then return end
+
+    for template_id, template in pairs(t.Templates) do
+        if template then
+            self.IntUpdatesPerTemplate[template_id] = self.IntUpdatesPerTemplate[template_id] or {}
+            local int_updates = self.IntUpdatesPerTemplate[template_id]
```
AmyJeanes: No need to set an int_updates variable here you can just set it directly to true e.g. `self.IntUpdatesPerTemplate[template_id][id] = true`

lua/tardis/settings/sh_miscellaneous.lua
```diff
@@ -247,6 +247,19 @@ TARDIS:AddSetting({
     name="NoInterior",
 })
 
+TARDIS:AddSetting({
+    id="legacy_door_type",
```
AmyJeanes, line 251: This can be a user personal preference and doesn't need to be a spawn and "TARDIS-wide" setting for everyone - is there any reason it was done this way in particular?

lua/tardis/sh_parts.lua
```diff
@@ -53,7 +53,19 @@ local overrides={
         local int=self.interior
         local ext=self.exterior
         if self._init and IsValid(int) and IsValid(ext) then
-            if (int:CallHook("ShouldThink")~=false) or (ext:DoorOpen() and self.ClientThinkOverride and LocalPlayer():GetPos():Distance(ext:GetPos())<TARDIS:GetSetting("portals-closedist")) or self.ExteriorPart then -- TODO: Improve
+            local think_ok = (int:CallHook("ShouldThink") ~= false)
+
+            local function is_visible_through_door()
+                local door = ext:DoorOpen()
```
AmyJeanes: This is much better for readability, but I am concerned all the local variable assignments may affect performance - would you mind checking to see if there is any noticeable drop in framerate for an interior with lots of parts from the original version? 

If so, we could still split it out onto multiple lines but maybe use comments instead to denote what's going on

lua/tardis/controls/sh_control_teleport.lua
```diff
@@ -14,7 +14,7 @@ TARDIS:AddControl({
                 if result then
                     TARDIS:Message(ply, "Controls.Teleport.Demat")
                 else
-                    if self:GetData("doorstatereal", false) then
+                    if self:GetData("doorstatereal", false) and not TARDIS:GetSetting("teleport-door-autoclose", self) then
```
AmyJeanes, line 17: This change itself is fine but it got me wondering why we're not just returning the reason for demat failure with another parameter returned from self:Demat and trying to work it out from the state. Maybe something for another issue/PR but thought I'd raise it

AmyJeanes (changes requested):
Couple of tiny things but seems to work well in-game, nice work!

lua/tardis/settings/sh_miscellaneous.lua
parar020100: It would be a mess when one can't open the doors and someone else can

Also, one of the use-cases I see is when portals are too laggy for someone (and they don't want that black replacement)

I think it's for the best for the TARDISes to be of certain "types"
It either has portals or a legacy-style entrance

parar020100: Everyone can spawn their own kind, and if you changed your mind you can just respawn it with a different type

AmyJeanes: Discussed on Discord, happy with the approach

lua/tardis/sh_parts.lua
parar020100: The performance isn't affected much for me, but I've made the condition a bit better

lua/tardis/controls/sh_control_teleport.lua
parar020100: #851 created

lua/tardis/sh_parts.lua
parar020100: 588f3e54ca30455fb4b173cca793071033e08ce4 should make it optimal

lua/tardis/interiors/modules/sh_templates.lua
parar020100: this was just a leftover from some previous code, removed in b0f79d581a3c8ce794415eabc8b959a0ab33bfe4

AmyJeanes (approved):
Nice work!

Issue #820, opened 2023-02-18
Option to disable animated icons for extensions 


PoogieTheGoose:

AmyJeanes:
Every extension would need to provide a non animated version of their icon, but if they did we could do this

parar020100:
#850 pretty much solved this

Issue #742, opened 2022-12-01
Spawn TARDISes using toolgun


parar020100:
This possibility was lost when spawn menu got all the interiors
It would be great to restore that

parar020100:
Done in #850

PR #850, opened 2023-03-12
Interior icons support

parar020100:
**Features:**
- "Interior" versions of the spawnmenu icons
  - By default, you can see the interior when hovering mouse over the icon
  - A setting allows to switch to interior icons only, which improves performance (#820)
  - Main addon has default icons at:
    `vgui/entities/tardis/default` and `vgui/entities/tardis/interiors/default`
    Extension creators can override default icons by putting their own into:
    `vgui/entities/tardis` and `vgui/entities/tardis/interiors`
- Added "spawn using toolgun" option (#742)

- Added a console variable `tardis2_spawnmenu_copy_id` which adds a "copy to clipboard" button
  - Instead of copying the entity class, the button copies the TARDIS ID
  - I thought this would be useful for developers while completely useless to others.
    The context menu of some TARDISes (with several versions) is already very big, so I've made a convar to enable it. We could probably add it to debug tools wiki page.

**Enhancements:**
- Adding a TARDIS to favorites no longer requires the game reload (works instantly)
- Added support for .jpg extension icons (for saving disk space)

**Internal changes & cleanup:**
- Refactoring of the spawnmenu module
  - Replaced some local functions into the ones in the TARDIS table
    (functions in the table are easier to debug since they update without a game reload)
- Converted old default icons from png to jpg format 
- Removed `Think` from the "favorite" button of context menu (for better performance)


i18n/languages/en.json
```diff
@@ -428,6 +428,8 @@
     "Settings.Sections.Misc.Spawning.RandomizeSkins.Description": "Whether or not TARDIS skin will be randomized when it's spawned",
     "Settings.Sections.Misc.Spawning.UseWinterSkins": "Use winter skins",
     "Settings.Sections.Misc.Spawning.UseWinterSkins.Description": "Whether or not winter TARDIS skins will be used while it's randomized",
+    "Settings.Sections.Misc.SpawnmenuInteriorIcons": "Show interiors in spawnmenu",
```
AmyJeanes: Clarify language here now that they show on hover normally to something like "Always show interiors in spawnmenu", same in description

lua/tardis/interiors/modules/sh_spawnmenu.lua
```diff
@@ -3,8 +3,18 @@ function TARDIS:SpawnByID(id)
     surface.PlaySound("ui/buttonclickrelease.wav")
 end
 
+TARDIS.InteriorIcons = {}
```
AmyJeanes, line 6: Random comment location as I can't comment on the pictures

The old default interior pic is very dark compared to the others 
image

AmyJeanes: image

lua/tardis/sh_interiors.lua
```diff
@@ -89,6 +91,13 @@ function TARDIS:SetupMetadata(id)
 end
 
 function TARDIS:CreateInteriorMetadata(id, ent)
+    if id == nil then
+        local cv_id = GetConVar("tardis2_create_tool_id"):GetString()
```
AmyJeanes: I think given how this works we should name the convar something a little more generic and then it could be used by other things outside the addon potentially too. Maybe `tardis2_default_interior` or similar

AmyJeanes (changes requested):
Looking pretty good! Some very minor notes but all working nicely in-game, nice work!

lua/tardis/sh_interiors.lua
parar020100: Renamed, I hope `tardis2_selected_interior` is fine

i18n/languages/en.json
parar020100: Changed

lua/tardis/interiors/modules/sh_spawnmenu.lua
parar020100: Done, + a bunch of improvements of other icons in 242b8ddcd5702b3e621f9efdcd13acb7009ce8b4

AmyJeanes (approved):
Looks good, nice work!

PR #859, opened 2023-03-23
TDG Update and e2 fixes

NovaAstral:
- Fixed E2's being able to cloak the TARDIS when it was out of power or artron
- Made E2's stop being able to use functions on someone elses TARDIS if they have Isomorphic enabled

-- Time Distortion Generator Changes:
- Added Wire support to the TDG with Activate and Radius inputs and Active, Radius, and Health outputs
- Increased max size of TDG to 2048 (Default is still 1000)
- Added custom radius support for the TDG between 0 and 2048 using wire inputs
- Gave the TDG health that can deplete wether its active or not.
- Slightly increased efficiency of the Think in the TDG (It will only velocity the tdg if its unfrozen)
- Moved the TDG activate timer into an actual timer instead of think


lua/entities/gmod_tardis/modules/sh_cloak.lua
```diff
@@ -50,7 +50,11 @@ if SERVER then
 
     ENT:AddHook("HandleE2", "cloak", function(self,name,e2)
         if name == "Phase" and TARDIS:CheckPP(e2.player, self) then
-            return self:ToggleCloak() and 1 or 0
+            if(self:GetData("power-state")) then
```
AmyJeanes: This could be simplified to `return (self:GetData("power-state", false") and self:ToggleCloak()) and 1 or 0`

(I think, worth testing!)

lua/entities/gmod_tardis/modules/sh_time_distortions.lua
```diff
@@ -1,12 +1,21 @@
...
 
-local search_radius = 1000
+local search_radius = 2048
 local explode_radius = 300
 
 local function TimeDistortionsPresent(pos, radius)
     for i,v in ipairs(ents.FindInSphere(pos, radius)) do
-        if v:GetClass() == "gmod_time_distortion_generator" and v:GetEnabled() then
-            return true
+        if(v:GetClass() == "gmod_time_distortion_generator" and v.On) then
```
AmyJeanes: Unless you have a good reason I think we can leave this to be `v:GetEnabled()` as we ideally shouldn't be directly accessing properties on other entities, same for the Radius should probably have a `GetRadius` function

lua/entities/gmod_tardis/modules/sh_time_distortions.lua
```diff
@@ -1,12 +1,21 @@
...
+            if(radius == search_radius) then
+                print(v:GetPos():Distance(pos))
```
AmyJeanes: Debugging print statement left in code

lua/entities/gmod_tardis/modules/sh_time_distortions.lua
```diff
@@ -1,12 +1,21 @@
...
+                if(v:GetPos():Distance(pos) <= v.Radius) then
+                    return true
+                end
+            else
+                if(v:GetPos():Distance(pos) <= radius) then
```
AmyJeanes: General comment about the if statements across all the changes they seem to have no space after the `if` and they have unnecessary brackets

lua/entities/gmod_tardis/modules/sh_time_distortions.lua
```diff
@@ -1,12 +1,21 @@
...
-local search_radius = 1000
+local search_radius = 2048
 local explode_radius = 300
 
 local function TimeDistortionsPresent(pos, radius)
     for i,v in ipairs(ents.FindInSphere(pos, radius)) do
-        if v:GetClass() == "gmod_time_distortion_generator" and v:GetEnabled() then
-            return true
+        if(v:GetClass() == "gmod_time_distortion_generator" and v.On) then
+            if(radius == search_radius) then
```
AmyJeanes: I think to avoid this weird check (this confused me for a solid few minutes until I worked out why it's here) it would be better to replace the radius parameter of this function with an `outside` boolean parameter and then you can reference the `search_radius` if it's true or the `ent.metadata.Interior.ExitDistance` when it's false. You'll probably also want to pass in the entity itself as the first parameter in order to access the metadata

lua/entities/gmod_time_distortion_generator/init.lua
```diff
@@ -18,89 +18,160 @@ function ENT:Initialize()
...
+
     if (phys:IsValid()) then
         phys:Wake()
     end
+
+    if not WireLib then 
+        return 
+    else
+        local inNames = {"Activate","Radius"}
+	    local inTypes = {"NORMAL","NORMAL"}
```
AmyJeanes: Formatting is a little off here

lua/entities/gmod_time_distortion_generator/init.lua
```diff
@@ -18,89 +18,160 @@ function ENT:Initialize()
...
 
+    self.Radius = 1000
+    self.EntHealth = 100
+    self.EntMaxHealth = 100
+
     if (phys:IsValid()) then
         phys:Wake()
     end
+
+    if not WireLib then
```
AmyJeanes: After the `return end` you don't need the `else` section you can just continue the code normally here

lua/entities/gmod_time_distortion_generator/init.lua
```diff
@@ -18,89 +18,160 @@ function ENT:Initialize()
...
+        if self.Broken then return end
+
+        if(value >= 1) then
+            if(self:GetEnabled() == false) then
+                self:TurnOn(true)
+            end
+        else
+            if(self:GetEnabled() == true) then
+                self:TurnOn(false)
+                Wire_TriggerOutput(self,"Active",0)
```
AmyJeanes: Is this `Wire_TriggerOutput` call needed? It looks like it's already called inside the TurnOn function

lua/entities/gmod_time_distortion_generator/init.lua
```diff
@@ -18,89 +18,160 @@ function ENT:Initialize()
...
+            end
+        end
+    elseif(iname == "Radius") then
+        self.Radius = math.Clamp(value,0,2048)
```
AmyJeanes: It would be good to have a `SetRadius` function similar to the TurnOn function to do this so it could be used by other things in future

lua/entities/gmod_time_distortion_generator/init.lua
```diff
@@ -18,89 +18,160 @@ function ENT:Initialize()
...
         self:SetColor(Color(255, 166, 0))
         phys:SetAngleVelocity(Vector(20, -20, -10))
         phys:EnableGravity(false)
         phys:SetVelocity(Vector(0, 0, 15))
 
         sound.Play("drmatt/tardis/seq_ok.wav", self:GetPos())
 
-        self.FlyTime = CurTime()
-        self.LastActivator = ply
+        timer.Create("tdg_enable_timer",3,1,function()
```
AmyJeanes: Usage of timers is dangerous in GMod as they are very error prone and unreliable, can we return this to using the CurTime + Think function method as before?

lua/entities/gmod_wire_expression2/core/custom/tardis2.lua
```diff
@@ -16,9 +16,13 @@ local function getTardis(ent)
     end
 end
 
-local function HandleE2(ent, name, ...)
+local function HandleE2(owner, ent, name, ...)
```
AmyJeanes: It would be good to rearrange this so that `self` was passed in as the first argument and then we could grab the `player` off that as well as pass it into `ent:HandleE2(name, self, ...)` like so, rather than passing in `self.player` and then `self` later on in the same call.

AmyJeanes: This is more of a comment for myself but we need to handle how this will interact with the legacy addon, currently there is effectively a copy of this file in the legacy addon but I'm thinking it might just be better to remove it and require the normal addon to use Wiremod against both legacy and the rewrite - otherwise it is a pain to keep updating legacy to keep it in sync with the rewrite for E2.

AmyJeanes (changes requested):
Hey Nova, many thanks for contributing! I like the general theme of your changes here and more wiremod support is always welcome, we have definitely been lacking in that department historically when adding new features!

I hope I'm not scaring you off with I believe your first contribution to this addon with my comments, I don't mean them in a bad way at all and it's really great to see someone else making changes to the core addon!

If you want to discuss any of them outside GitHub you can feel free to message me on Discord, you already know my username as you asked me about how to create a branch earlier.

I know @francodml in particular will be pleased to see further Wiremod support!

Again nice work, and for a first contribution especially this is great stuff!

lua/entities/gmod_wire_expression2/core/custom/tardis2.lua
AmyJeanes: On further thinking actually I think it might be best left to the HandleE2 function to deal with this which already has the player passed in as part of the E2 context.

I think the GetX functions are okay to run against any TARDIS regardless of isomorphic for example which we can only really handle in that function. There is an `ENT:CheckSecurity(ply)` function you can use to check permissions for this for things like `SetFlight` etc

lua/entities/gmod_tardis/modules/sh_time_distortions.lua
NovaAstral: I don't have a good reason, I accidentally'd the GetEnabled and only noticed right before commit

lua/entities/gmod_tardis/modules/sh_time_distortions.lua
NovaAstral: I personally find it easier to read this way, and it just feels better (probably because I started coding as an E2 user) but I can change it if you'd like

AmyJeanes: If you could please - thanks, then it matches Lua conventions and the conventions of the TARDIS codebase

lua/entities/gmod_tardis/modules/sh_time_distortions.lua
NovaAstral: I'm not quite sure how to do this

lua/entities/gmod_tardis/modules/sh_time_distortions.lua
AmyJeanes: Fixed

lua/entities/gmod_tardis/modules/sh_time_distortions.lua
AmyJeanes: Fixed

lua/entities/gmod_tardis/modules/sh_time_distortions.lua
AmyJeanes: As discussed on Discord I've done this bit for you

lua/entities/gmod_time_distortion_generator/init.lua
```diff
@@ -18,59 +18,129 @@ function ENT:Initialize()
...
+    self.EntMaxHealth = 100
+
+    if phys:IsValid() then
         phys:Wake()
     end
+
+    if WireLib then
+        local inNames = {"Activate","Radius"}
+        local inTypes = {"NORMAL","NORMAL"}
+	WireLib.CreateSpecialInputs( self,inNames,inTypes)
```
AmyJeanes: Formatting is a little off here, looks like you've got a tab character instead of spaces and there's a space after the left bracket

lua/entities/gmod_tardis/modules/sh_cloak.lua
```diff
@@ -50,7 +50,7 @@ if SERVER then
 
     ENT:AddHook("HandleE2", "cloak", function(self,name,e2)
         if name == "Phase" and TARDIS:CheckPP(e2.player, self) then
-            return self:ToggleCloak() and 1 or 0
+                return (self:GetData("power-state", false) and self:ToggleCloak()) and 1 or 0
```
AmyJeanes: You should use `self:GetPower()` here instead of checking the power state directly

lua/entities/gmod_time_distortion_generator/init.lua
```diff
@@ -18,59 +18,129 @@ function ENT:Initialize()
...
+        Wire_CreateOutputs(self,{"Active","Radius","Health"})
+
+        Wire_TriggerOutput(self,"Radius",1000)
```
AmyJeanes: This should probably call it with self.Radius instead of hardcoding to 1000

lua/entities/gmod_wire_expression2/core/custom/tardis2.lua
```diff
@@ -129,79 +133,79 @@ end
 ]]
 
 e2function string entity:tardisGetData(string id)
-    return tostring(this:GetData(id,false))
+    return tostring(self.player, this:GetData(id,false))
```
AmyJeanes: We probably don't want `self.player` on this one

lua/entities/gmod_wire_expression2/core/custom/tardis2.lua
```diff
@@ -16,11 +16,15 @@ local function getTardis(ent)
     end
 end
 
-local function HandleE2(ent, name, ...)
+local function HandleE2(ent, type, name, e2, ...)
     if IsValid(getTardis(ent)) then
-        return ent:HandleE2(name, ...)
+        if type == "Setter" and ent:CheckSecurity(e2.player) == false then
+            TARDIS:ErrorMessage(e2.player,"Security.E2.UseDenied")
```
AmyJeanes: How does this behave in E2 if it's expected to get something back but returns nothing now - does it cause any issues?

i18n/languages/en.json
```diff
@@ -215,6 +215,7 @@
     "Destination.LockedDemat": "Destination locked, dematerialising...",
     "Destination.LockedReadyToDemat": "Destination locked, ready to dematerialise",
     "Destination.LockedReadyToMat": "Destination locked, ready to materialise",
+    "E2.Error": "Can't call TARDIS functions on something other than a TARDIS.",
```
AmyJeanes: Could we change the new strings to say Expression2 rather than E2 for clarity?

AmyJeanes (changes requested):
This is looking great and almost perfect! Just a couple of small things to sort out before we can merge :)

lua/entities/gmod_wire_expression2/core/custom/tardis2.lua
AmyJeanes: Looks good!

lua/entities/gmod_wire_expression2/core/custom/tardis2.lua
NovaAstral: I don't actually even know how that happened

lua/entities/gmod_wire_expression2/core/custom/tardis2.lua
NovaAstral: From what I tested, the E2 just does nothing

AmyJeanes (approved):
LGTM! Nice work 🙂

lua/entities/gmod_wire_expression2/core/custom/tardis2.lua
AmyJeanes: I'll handle this before deployment

PR #862, opened 2023-03-26
Fixed #772 - cloak not always playing animation

### LINKED ISSUE #772: Cloak doesn't always play animation
> Theres a small chance for the mechanism to skip the animation and just instantly snap into the cloaked position. This happens both ways with cloaking and uncloaking.

parar020100:
Closes #772


AmyJeanes (approved):
Nice!

PR #871, opened 2023-03-29
Added door functions to E2.

noatblok:
Added E2 functions to check if the doors are open, and to toggle the doors.

### DIFF
--- a/lua/entities/gmod_tardis/modules/sh_doors.lua
+++ b/lua/entities/gmod_tardis/modules/sh_doors.lua
@@ -85,7 +85,7 @@ if SERVER then
 
     ENT:AddHook("HandleE2", "door", function(self,name,e2)
         if name == "Door" and TARDIS:CheckPP(e2.player, self) then
-            return self:ToggleDoor()
+            return self:ToggleDoor() and 1 or 0
         elseif name == "GetDoor" then
             return self:DoorOpen(false) and 1 or 0
         end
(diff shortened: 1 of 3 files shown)


lua/entities/gmod_tardis/modules/sh_doors.lua
```diff
@@ -83,6 +83,14 @@ if SERVER then
         end
     end
 
+    ENT:AddHook("HandleE2", "door", function(self,name,e2)
+        if name == "Door" then
+            return self:ToggleDoor()
+        elseif name == "GetDoor" and TARDIS:CheckPP(e2.player, self) then
```
AmyJeanes: We need to move the CheckPP call up to the ToggleDoor if statement rather than this one

AmyJeanes (changes requested):
Hey, thanks for contributing! Just a small little fix needed I think and then it should be good to go 🙂

lua/entities/gmod_tardis/modules/sh_doors.lua
noatblok: Should be fixed.

noatblok:
Updated the E2 description, aswell as having tardisDoor respond with a 1 or a 0 depending on the outcome.

AmyJeanes (approved):
Looks good!

I've made a small tweak to use the real door state instead of the estimated one which for the purposes of E2 usage makes more sense so you can have a script like this:

```lua
local Success=TARDIS:tardisDoor()
if (Success==1){
    if(TARDIS:tardisDoorOpen()){
        print("TARDIS door open")
    }else{
        print("TARDIS door closed")
    }
}
```

Otherwise it would say the door is open even if it is closing

PR #873, opened 2023-03-31
Make E2 Functions also move the console controls

noatblok:
Added a new function called `TARDIS:UsePart`, when called, it plays the control animation alongside the sound. Added `UsePart` to the `HandleE2` events.

### DIFF
--- a/lua/entities/gmod_tardis/modules/sh_cloak.lua
+++ b/lua/entities/gmod_tardis/modules/sh_cloak.lua
@@ -42,7 +42,7 @@ if SERVER then
         self:CallHook("CloakToggled", on)
         return true
     end
-    
+
     function ENT:ToggleCloak()
         local on = not self:GetData("cloak", false)
         return self:SetCloak(on)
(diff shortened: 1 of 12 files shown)


parar020100:
I have a slight issue with the way this function works
`TARDIS:GetPartByAction(ent, action)`

Seems like it presumes there's only one part per action
Right now many TARDISes have several parts with same controls
Is a random one gonna be selected in this?

noatblok:
I haven't actually tested that, at the moment it just looks through the interior controls, and finds the first one that triggers the corresponding action (HADS, Cloak, etc.). I'm not sure what the best way to resolve this is, my only idea is to add some sort of argument to let the E2 code choose which control (If there are more than one).

parar020100:
I can also see `TARDIS:UsePart(part)` cause potential issues with `PART:Use()` functions specified.
For instance, that function may specify the delays and additional effects & part interactions

I think we should stick to #885 and make extension creators choose which parts should be toggled.
Not finding and moving them automatically.

AmyJeanes:
Hey, due to inactivity and the fact this PR would need significant work to be accepted, I am going to close this for now. If you pick this up again please feel free to raise a new PR.

PR #874, opened 2023-03-31
Chameleon circuit

parar020100:
This feature is so good that I propose we release it ASAP


AmyJeanes (approved):
LGTM! People are going to love this!

PR #877, opened 2023-04-01
Poogie nextbots

PoogieTheGoose:
Made nextbots and props teleport inside when landing on them. Got an issue with ragdolls though, so i leaved the "classes"
table to make them in future.


AmyJeanes (approved):
Nice! Good usage of a lookup table as well for performance

PR #895, opened 2023-04-15
Disable console control + close tips option

PoogieTheGoose:
1)Added a control that allows to disable all other controls(except for the throttle). Made mostly for roleplay and stuff.
2)Added an option to make tips only appear if you look at the control they attached to. Made for the extensions with lots of controls close to each other so its pretty hard to navigate.


AmyJeanes (changes requested):
All looking good to me in-game and the code as well, nice work!

We just need to add an icon to the built in themes for the new Toggle Console control and then I'm happy to merge

image

AmyJeanes (approved):

PR #911, opened 2023-05-25
New exterior light size option

PoogieTheGoose:
Added a new option for the exterior light, so you can adjust it to your preference. Personally i love when the TARDIS lights up in the distance like the Legacy ones did. They were really cozy and pleasant even in the distance. Im not sure with the default value though. Thats up to your choice. Im probably gonna be away for a month or so and its gonna be a bit problematic to code anything.


AmyJeanes (approved):
Nice! Just made a couple little tweaks but a useful option, especially for RP

PR #914, opened 2023-05-30
Demat sound added

PoogieTheGoose:
I thought if we have mat_fast sound when using fullflight why dont we have demat_fast sound as well. It would be a nice addition for extensions with different teleport sequence speed for fast flight.


AmyJeanes (approved):
Nice little tweak for extension devs! I had to make a little fix to it as it would have always used the fast sound but otherwise looks good.

PR #916, opened 2023-06-08
Warning state

### LINKED ISSUE #597: Enable warning-state (red lights) when HADS is activated

PoogieTheGoose:
Made warning state work not only when health goes low but also with other scenarios. For example hads trigger 597 . Can be used in custom hooks in other extensions as well.


AmyJeanes (approved):
Looks good, nice! And good eye for performance too by not updating it every tick.

I made a very tiny change to use proper Lua syntax as != is 'Garry Lua'

PR #917, opened 2023-06-15
Vertical landing and multiple tweaks

parar020100:
Description: <TBA>

This should be ready for review
However, I'd want some testing for vertical landing before it's merged

Contains many important fixes (broken redecoration, tips, insta-death by physgun, hidden interiors spawning etc.)

I'll link the related issues soon.


lua/entities/gmod_tardis/modules/sh_falling.lua
```diff
@@ -0,0 +1,148 @@
...
+        if self:GetData("free_movement")
+            and TARDIS:GetSetting("flight-externalsound")
+            and TARDIS:GetSetting("sound")
+        then
+            local p=math.Clamp(self:GetVelocity():Length()/125,0,15)
+            if self.fallsound and self.fallsound:IsPlaying() then
+                local ply=LocalPlayer()
+                local e=ply:GetViewEntity()
+                if not IsValid(e) then e=ply end
+                if e:EntIndex()==-1 then -- clientside prop
```
AmyJeanes, line 112: Why do we need the clientside prop checks here or anything outside of `ply:GetTardisData("outside" / "exterior")?

I noticed that even in third person it still uses the doppler version of the effect and to my understanding it should just be the normal one in that scenario, similar to the flight sound in third person

lua/entities/gmod_tardis/modules/sh_falling.lua
```diff
@@ -0,0 +1,148 @@
...
+                    local doppler = (pos:Distance(spos+e:GetVelocity())-pos:Distance(spos+self:GetVelocity()))/200
+                    self.fallsound:ChangePitch(math.Clamp(95+p+doppler,80,120),0.1)
+                end
+
+                self.fallsound:ChangeVolume(0.75 * p / 20)
+            else
+                self.fallsound = CreateSound(self, self.metadata.Exterior.Sounds.FlightFallWind)
+                self.fallsound:SetSoundLevel(90)
+                self.fallsound:ChangeVolume(0.75 * p / 20)
+                self.fallsound:Play()
```
AmyJeanes, line 134: Letting go of a TARDIS after dragging it around at speed will cause it to make a pop sound for some reason

lua/entities/gmod_tardis/modules/sh_falling.lua
AmyJeanes, line 1: Landing in the water does not work properly, it should just ignore water if it hits it

AmyJeanes, line 1: High speed horizontal lands look very strange, we should restrict the maximum horizontal velocity that this can work at

AmyJeanes, line 1: The landing sound can play even after the smallest of drops, we should increase the threshold for this a little

AmyJeanes, line 1: The landing sound can go off multiple times in quick succession if the landing doesn't stick properly the first time, we should have a cooldown on this for a second or two.

lua/entities/gmod_tardis/modules/sh_health.lua
```diff
@@ -241,6 +241,10 @@ if SERVER then
     end)
 
     ENT:AddHook("PhysicsCollide", "Health", function(self, data, collider)
+        if self:IsPlayerHolding() then
+            self:SetVelocity(Vector(0,0,0))
+            self:ForcePlayerDrop()
```
AmyJeanes: I'm not sure if dropping it is a good idea, perhaps we should just ignore collisions above a really high speed when the player is holding the TARDIS? I still think you should be able to move it around with the physgun we just want to avoid the physics engine glitching out and registering extremely high speed hits by a player forcing it into the ground / wall / etc

lua/entities/gmod_tardis_interior/modules/cl_tips.lua
```diff
@@ -179,15 +179,15 @@ hook.Add("HUDPaint", "TARDIS-DrawTips", function()
         end
 
         local part = tip.part and interior:GetPart(tip.part)
-        local partok = IsValid(part)
-        local shoulddraw = TARDIS:GetSetting("tips_show_all") or tip:GetHighlight() or (partok and part:BeingLookedAtByLocalPlayer())
+        local shoulddraw =  tip:GetHighlight() or TARDIS:GetSetting("tips_show_all")
```
AmyJeanes: nit: extra space here

lua/entities/gmod_tardis/modules/sh_falling.lua
AmyJeanes, line 1: I think we should have some kind of control or option here, or perhaps both e.g. default and control for ad hoc changes similar to isomorphic on by default setting. I can see people wanting the old behaviour here, or even switching between it on the fly.

AmyJeanes (changes requested):
Very interesting feature, works pretty well!

lua/entities/gmod_tardis/modules/sh_redecorate.lua
```diff
@@ -175,7 +175,7 @@ if SERVER then
             parent:SetData("redecorate_mat_started", true)
```
AmyJeanes: Error when redecorating, not sure if related to this PR but worth fixing anyway

```
[tardis] addons/tardis/lua/entities/gmod_tardis/modules/sh_redecorate.lua:174: Tried to use a NULL entity!
  1. ForcePlayerDrop - [C]:-1
   2. func - addons/tardis/lua/entities/gmod_tardis/modules/sh_redecorate.lua:174
    3. v - addons/tardis/lua/entities/gmod_tardis/modules/sh_timer.lua:28
     4. CallHook - addons/tardis/lua/entities/gmod_tardis/shared.lua:70
      5. unknown - addons/doors/lua/entities/gmod_door_exterior/init.lua:91
```

lua/entities/gmod_tardis/modules/sh_falling.lua
parar020100: To be honest I'm not very sure about how the doppler sound is to work. 
My intention was to make the wind sound change its pitch as the speed increases - and since the flight sound works realistically, I just kept everything. Should this part not be there?

lua/entities/gmod_tardis/modules/sh_redecorate.lua
parar020100: done in e826d17c0b7c131dabe4c67f47e7f653b5bbee9b

lua/entities/gmod_tardis/modules/sh_health.lua
parar020100: ef817b1867f945cae0e218fe09299b66f6e699e4
this made a much better fix

lua/entities/gmod_tardis_interior/modules/cl_tips.lua
parar020100: c9969dc0581c4bda6cf17786877a6874a37eeb40

PR #924, opened 2023-09-20
Unlock sound

PoogieTheGoose:
This should be a thing for a long time now. Having one sound per lock and unlock really makes a lot of "Did i lock the door or unlocked it right now?" moments. Exterior sound is also provided.


AmyJeanes (approved):
Thanks!

PR #925, opened 2023-09-21
Fullflight seprate delay

PoogieTheGoose:
Speaks for itself. The time for demat delay doesnt compensate for mat delay so this feature is just necessary for people who use the delay feature(me)


AmyJeanes (approved):
Thanks!

PR #928, opened 2023-10-07
Bug fixing from #917

### LINKED ISSUE #853: Hidden interiors like toyota_base can spawn in redecoration

### LINKED ISSUE #887: The player can still pick the TARDIS with HADS activated
This shouldn't happen

parar020100:
Since the vertical falling feature is not ready yet, I'd suggest these changes to be merged first.
They fix a whole bunch of problems, as described by the commits.
Perhaps a minor release could be in place?

P.S. most of these changes have been reviewed in #917
I've fixed comments to that PR (apart from the falling)


lua/tardis/controls/sh_control_float.lua
```diff
@@ -8,7 +8,7 @@ TARDIS:AddControl({
         end
     end,
     serveronly=true,
-    power_independent = false,
+    power_independent = true,
```
AmyJeanes, line 11: Not sure I get why we changed this because it will still fail with power off anyway just with a different message?

lua/entities/gmod_tardis/modules/sh_health.lua
```diff
@@ -15,33 +15,54 @@ ENT:AddHook("SettingChanged","maxhealth-changed", function(self, id, val)
...
+    self:CallCommonHook("OnHealthChange", new_health, old_health)
+    if new_health == 0 and old_health ~= 0 then
+        self:CallCommonHook("OnHealthDepleted")
     end
-    self:SetData("health-val", newhealth, true)
-    self:CallCommonHook("OnHealthChange", newhealth, oldhealth)
+end
+
+if SERVER then
+    ENT:OnMessage("health_change", function(self, data, ply)
```
AmyJeanes: This doesn't appear to have any kind of security around it and anyone running clientside scripts could change the TARDIS health at any time

lua/entities/gmod_tardis/modules/sh_health.lua
```diff
@@ -15,33 +15,54 @@ ENT:AddHook("SettingChanged","maxhealth-changed", function(self, id, val)
...
 
 function ENT:GetHealth()
     return self:GetData("health-val", 0)
 end
 
+function ENT:IsAlive()
+    return (self:GetHealth() ~= 0)
+end
+
+function ENT:GetRepairPrimed()
```
AmyJeanes, line 48: This code is duplicated from fixes already made in dev for both `GetRepairPrimed` and `GetRepairing`

lua/entities/gmod_tardis/modules/sv_handbrake.lua
```diff
@@ -28,9 +28,9 @@ ENT:AddHook("ShouldFailDemat", "handbrake", function(self, force)
 end)
 
 ENT:AddHook("HandbrakeToggled", "vortex", function(self, on)
-    if on and self:GetData("teleport") or self:GetData("vortex") then
+    if on and (self:GetData("teleport") or self:GetData("vortex")) then
```
AmyJeanes: Would be easier to read if we return early `if not on` rather than this

lua/tardis/controls/sh_control_teleport.lua
```diff
@@ -27,7 +27,7 @@ TARDIS:AddControl({
     end,
     serveronly=true,
     power_independent = false,
-    bypass_console_toggle = true,
+    bypass_console_toggle = false,
```
AmyJeanes: Might as well just remove these lines rather than having them to false on this and the double teleport control

lua/entities/gmod_tardis/modules/teleport/sh_tp_vortex.lua
```diff
@@ -23,7 +23,7 @@ function ENT:ToggleFastRemat()
 end
 
 function ENT:SetFastRemat(on, force)
-    if (not force) and (self:CallHook("CanToggleFastRemat") == false) then
+    if self:CallHook("CanToggleFastRemat", force) == false then
```
AmyJeanes, line 26: These changes are not required as redecoration was fixed in a different way already

AmyJeanes (changes requested):

lua/tardis/controls/sh_control_float.lua
parar020100: I haven't found any reasons for it to fail in the code
Will test though

The point of the change is to allow the player to **disable** anti-gravs when powered-down

So if you mean the `CanTurnOnFloat` hook, that's okay since there's no `CanTurnOffFloat` hook

lua/entities/gmod_tardis/modules/sh_health.lua
parar020100: right
I think we need to make `ChangeHealth` completely serverside, idk why it was being called from a client anyway

lua/entities/gmod_tardis/modules/sh_health.lua
parar020100: Rebase conflict, my bad

lua/entities/gmod_tardis/modules/teleport/sh_tp_vortex.lua
parar020100: where and how was it fixed?

lua/tardis/controls/sh_control_float.lua
parar020100: **UPD**: double-checked, you can disable anti-gravs when power is disabled
that was the goal
not turning on but turning off (so that you don't get stuck)

lua/tardis/controls/sh_control_teleport.lua
parar020100: b70f996f3de524374052efcd1b117246cdb24335

lua/entities/gmod_tardis/modules/sh_health.lua
parar020100: 9115edec229fd2a660e9f7d8946ec6f94d1ef27f

lua/entities/gmod_tardis/modules/sv_handbrake.lua
parar020100: makes sense
8f56cf4a5a6392263641bc9f306fb854932b7440

lua/entities/gmod_tardis/modules/sh_health.lua
parar020100: done
9115edec229fd2a660e9f7d8946ec6f94d1ef27f

lua/entities/gmod_tardis/modules/teleport/sh_tp_vortex.lua
parar020100: I believe it is better within the `CenToggleFastRemat` hook since even though we are restricting it for teleport/vortex, there might be other restrictions that don't even apply to force fast remat toggle

AmyJeanes: Makes sense to me, no harm in doing this just thought worth noting

lua/tardis/controls/sh_control_float.lua
AmyJeanes: Ah okay gotcha, didn't realise that was the reason I just tried turning it on

AmyJeanes (approved):
Thank you! Nice work

PR #929, opened 2023-10-08
Hads speed

PoogieTheGoose:
Made HADS take off a bit quicker(1,3*FullFlight Speed so people dont need to add anything in their addons) , so a few RPG shots wont be enough to stop it, making HADS a bit less useless in PvP situations.


AmyJeanes (changes requested):
Hmm, I'm not so sure if we should do this - I don't like hardcoding a 1.3 multiplier and the fact that it will cause sound mismatches when HADS is activated isn't great either.

I don't mind if an option was added but I think it should be opt in for each interior and with the ability to choose a new sound to go with it

PoogieTheGoose:
In the show itself, hads demat had ordinary demat sounds so i cant suggest anything about that.
I'll readd the sequence speed for the HADS though

PoogieTheGoose:
Done

AmyJeanes (changes requested):
Okay that's looking a lot better now, but the sound ends quite abruptly and doesn't have the post demat zappy sound effects if you know what I mean - if we could get another version of the sound with that bit on the end I'll be happy with this!

parar020100:
Me and Poogie are working on this on a different branch, so I'm gonna close the PR

PR #932, opened 2023-10-14
Blocked exit

Evgeni420:
- block doors when it's impossible to exit
- added `SlowThink` hooks serverside

### DIFF
--- a/lua/entities/gmod_tardis/init.lua
+++ b/lua/entities/gmod_tardis/init.lua
@@ -45,6 +45,7 @@ end
 function ENT:PhysicsCollide(colData, collider)
     self:CallHook("PhysicsCollide", colData, collider)
 end
+
 function ENT:OnTakeDamage(dmginfo)
     if self:CallHook("ShouldTakeDamage",dmginfo)==false then return end
     self:CallHook("OnTakeDamage", dmginfo)
@@ -56,3 +57,13 @@ duplicator.RegisterEntityClass("gmod_tardis", function(ply, data)
     ent:Initialize()
     return ent
 end, "Data")
+
+function ENT:Think()
+    self.BaseClass.Think(self)
+    if not self._init then return end
+
+    if CurTime() >= (self.nextslowthink or 0) then
+        self.nextslowthink = CurTime() + 1
+        self:CallHook("SlowThink")
+    end
+end
--- a/lua/entities/gmod_tardis/init.lua
+++ b/lua/entities/gmod_tardis/init.lua
@@ -45,7 +45,6 @@ end
 function ENT:PhysicsCollide(colData, collider)
     self:CallHook("PhysicsCollide", colData, collider)
 end
-
 function ENT:OnTakeDamage(dmginfo)
     if self:CallHook("ShouldTakeDamage",dmginfo)==false then return end
     self:CallHook("OnTakeDamage", dmginfo)
@@ -57,13 +56,3 @@ duplicator.RegisterEntityClass("gmod_tardis", function(ply, data)
     ent:Initialize()
     return ent
 end, "Data")
-
-function ENT:Think()
-    self.BaseClass.Think(self)
-    if not self._init then return end
-
-    if CurTime() >= (self.nextslowthink or 0) then
-        self.nextslowthink = CurTime() + 1
-        self:CallHook("SlowThink")
-    end
-end
--- a/lua/entities/gmod_tardis/modules/sh_doors.lua
+++ b/lua/entities/gmod_tardis/modules/sh_doors.lua
@@ -246,12 +246,11 @@ if SERVER then
     ENT:AddHook("DoorCollisionOverride", "doors", function(self)
         local td = {}
         local ply = self:GetCreator()
-        td.mins = ply:OBBMins()
-        td.mins.z = td.mins.z * 0.5
-        td.maxs = ply:OBBMaxs()
-        td.maxs.z = td.maxs.z * 0.5
+        td.mins = ply:OBBMins() * 0.5
+        td.maxs = ply:OBBMaxs() * 0.5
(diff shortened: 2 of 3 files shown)


lua/entities/gmod_tardis/init.lua
```diff
@@ -42,7 +42,18 @@ end
...
 function ENT:OnTakeDamage(dmginfo)
     if self:CallHook("ShouldTakeDamage",dmginfo)==false then return end
     self:CallHook("OnTakeDamage", dmginfo)
+end
+
+function ENT:Think()
+    self.BaseClass.Think(self)
+    if not self._init then return end
+
+    if CurTime() >= (self.nextslowthink or 0) then
```
AmyJeanes: This is good but it should be implemented into the Doors addon instead as it's more of a base code feature, that's also where the fast think is implemented so it makes sense to be in there too.

I've added you as a collaborator to https://github.com/MattJeanes/Doors now so you can do this

lua/entities/gmod_tardis/modules/sh_doors.lua
```diff
@@ -147,18 +135,32 @@ if SERVER then
...
+            self:UpdateDoorCollision()
         end
     end)
 
+    function ENT:UpdateDoorCollision()
+        local open = self:DoorOpen(true)
+        local override = self:CallHook("DoorCollisionOverride")
+
+        local door_ext = TARDIS:GetPart(self,"door")
+        local door_int=TARDIS:GetPart(self.interior,"door")
```
AmyJeanes: nitpick: inconsistent spacing between these lines

lua/entities/gmod_tardis/modules/sh_doors.lua
```diff
@@ -147,18 +135,32 @@ if SERVER then
...
+
+        if (override == nil and open) or override==false then
+            if IsValid(door_ext) then door_ext:SetSolid(SOLID_NONE) end
+            if IsValid(door_int) then door_int:SetCollisionGroup( COLLISION_GROUP_WORLD ) end
+        elseif override or override==nil then
```
AmyJeanes: Could this be replaced with `elseif override ~= false`?

lua/entities/gmod_tardis/modules/sh_doors.lua
```diff
@@ -147,18 +135,32 @@ if SERVER then
...
 
+    function ENT:UpdateDoorCollision()
+        local open = self:DoorOpen(true)
+        local override = self:CallHook("DoorCollisionOverride")
+
+        local door_ext = TARDIS:GetPart(self,"door")
+        local door_int=TARDIS:GetPart(self.interior,"door")
+
+        if (override == nil and open) or override==false then
+            if IsValid(door_ext) then door_ext:SetSolid(SOLID_NONE) end
```
AmyJeanes: nitpick: inconsistent spacing around the brackets on `SetSolid` on these lines

lua/entities/gmod_tardis_interior/init.lua
```diff
@@ -30,4 +30,14 @@ end
 function ENT:OnTakeDamage(dmginfo)
     if self:CallHook("ShouldTakeDamage",dmginfo)==false then return end
     self:CallHook("OnTakeDamage", dmginfo)
-end
\ No newline at end of file
+end
+
+function ENT:Think()
```
AmyJeanes: Should be in Doors addon as per the other comment :)

lua/entities/gmod_tardis/modules/sh_doors.lua
```diff
@@ -236,6 +238,29 @@ if SERVER then
...
+        local td = {}
+        local ply = self:GetCreator()
+        td.mins = ply:OBBMins()
+        td.mins.z = td.mins.z * 0.5
+        td.maxs = ply:OBBMaxs()
+        td.maxs.z = td.maxs.z * 0.5
+        td.mask = MASK_NPCWORLDSTATIC
+        local fallback = self:LocalToWorld(self.Fallback.pos)
+
+        local z_ops = {0, 0.2, 0.4, 0.5, 0.6, 0.8, 1}
```
AmyJeanes, line 255: This might be slightly too pessimistic as it is possible to place the TARDIS in front of a wall in a place that you should be able to walk in through the side and similarly walk out but it blocks both the entrance and exit.

Here's an example from `freespace_revolution` on the default interior where it blocks me but I should be able to get in/out:

image

AmyJeanes (changes requested):
This is a great quality of life feature and really helps with the finishing touches on the addon. Thanks for working on this, it's a fantastic first contribution.

I've got a couple of small things I'd like some tweaks on but it's all minor stuff. Great work!

lua/entities/gmod_tardis/modules/sh_doors.lua
parar020100: Should we make a smaller scanning area?

AmyJeanes: Also wondering if this is open to abuse in MP, can you just stand in front of the door to block people getting out for example?

parar020100: > Also wondering if this is open to abuse in MP, can you just stand in front of the door to block people getting out for example?

Not abusable
See line `td.mask = MASK_NPCWORLDSTATIC` - it only filters the world

As for the exit condition
We might make the scanning box smaller but in general testing it seems to be forbidding exit when the player would get stuck in the textures

AmyJeanes: Yeah I think it just needs to be made a very slight bit smaller otherwise seems good!

parar020100:
We've decided to put this on hold for a while

PR #934, opened 2023-10-15
Navigation improvements

### LINKED ISSUE #246: Easier rotation for destination control
Maybe a button to toggle between movement and rotation might be a nice idea, should fix all the weirdness in the controls we have at the moment. If anyone has any ideas do let me know in the comments.

### LINKED ISSUE #547: Random Location Terrain Issue
When using Random Location (ground) Destination travel on maps with terrain such as hills and cliffs, often it is possible for the Tardis to land on these inclines, however it does not snap to them and results in the Tardis falling, resulting in damage depending on the fall
Tardis
.
Tardis 2

### LINKED ISSUE #668: [control] "Keep upright"
From Discord suggestion by TheStoneWolf:
> the Tardis should have a "keep upright" control so it can drop from high up and still land properly

### LINKED ISSUE #732: Improve grounded random coordinates selection algorithm
Right now, most random coordinates end up at the highest level of the map
The algorithm gets a random flight position and uses "snap to ground" feature

New proposal:
- have several modes selected randomly
- one of them will be the same as current (snap to floor)
- one will randomize heights after randomizing overall position
- one will prioritize lower positions (so that you can get into rooms)
- one will select a random position from the saved positions list

parar020100:
### Changelog:
- made the TARDIS land normally after a vertical fall (instead of bouncing around like a tennis ball)
  - a drum sound after normal landing and a crash sound after falling with a high speed (from NOTD)
  - when the TARDIS lands vertically, less collision damage is taken
  - while falling / moving without flight or anti-gravs, press CTRL / "flight-down" key to align the TARDIS vertically
  - added a wind sound when the TARDIS is free falling (the volume is proportional to the falling time)
  
... (TBA)

Closes #246, #547

### TODO:
- [x] Make falling alignment manual
- [x] Make it impossible to use the align control while staying on the ground
- [x] Make this fix #547 (stop the TARDIS from getting stuck in textures)
- [x] Make snap to floor ignore NPCs and players
- [x] Rotation selection in destination mode
- [x] Fix drum playing after redecoration


lua/entities/gmod_tardis/modules/sh_destination.lua
```diff
@@ -92,22 +100,83 @@ TARDIS:AddKeyBind("destination-demat",{
...
     name="SnapToFloor",
     section="Destination",
     func=function(self,down,ply)
-        if ply:GetTardisData("destination") then
+        if down and ply:GetTardisData("destination") then
             local prop = self:GetData("destinationprop")
             if IsValid(prop) then
-                local pos = util.QuickTrace(prop:GetPos()+Vector(0,0,50), Vector(0,0,-1)*99999999).HitPos
+                local prevpos = self:GetData("destination_snaptofloor_lastpos")
+                local should_glue = (prevpos == prop:GetPos())
```
AmyJeanes: In playtesting I found there were several places like here on `freespace_revolution` where this seems to work very inconsistently
image

lua/entities/gmod_tardis/modules/sh_destination.lua
```diff
@@ -92,22 +100,83 @@ TARDIS:AddKeyBind("destination-demat",{
...
+TARDIS:AddKeyBind("destination-find-random",{
+    name="FindRandom",
+    section="Destination",
+    func=function(self,down,ply)
+        if down and ply:GetTardisData("destination") then
+            local prop = self:GetData("destinationprop")
+            if IsValid(prop) then
+                prop:SetAngles(Angle(0,prop:GetAngles().y,0))
+                local grounded = not self:GetData("destination_random_grounded")
+                self:SetData("destination_random_grounded", grounded)
```
AmyJeanes: Is this intended to flip through a sky pos and ground pos on every press? Feels a little weird

lua/entities/gmod_tardis/modules/sh_destination.lua
```diff
@@ -92,22 +100,83 @@ TARDIS:AddKeyBind("destination-demat",{
...
     func=function(self,down,ply)
-        if ply:GetTardisData("destination") then
+        if down and ply:GetTardisData("destination") then
             local prop = self:GetData("destinationprop")
             if IsValid(prop) then
-                local pos = util.QuickTrace(prop:GetPos()+Vector(0,0,50), Vector(0,0,-1)*99999999).HitPos
+                local prevpos = self:GetData("destination_snaptofloor_lastpos")
+                local should_glue = (prevpos == prop:GetPos())
+
+                local pos, ang = self:GetGroundedPos(prop:GetPos(), should_glue)
```
AmyJeanes: This position does not ignore the TARDIS itself like the right click does so you can snap on top of itself 
image

lua/entities/gmod_tardis/modules/sh_falling.lua
```diff
@@ -0,0 +1,191 @@
+if SERVER then
+    function ENT:IsVerticalLanding(collision_data)
+        local ang = self:GetAngles()
+        local angmax = math.max(math.abs(ang.p), math.abs(ang.r))
+
+        return (angmax < 30 and collision_data.HitNormal.z < -0.85 and collision_data.OurOldVelocity.z < 0)
```
AmyJeanes: I think we need to take into account horizontal velocity here also as you can do a ridiculously fast landing and just stop instantly

AmyJeanes: Also if there is a reasonable amount of angular velocity because if you end up in a spin or doing flips or something and just so happen to hit the ground well enough it will also instantly stop which feels wrong

AmyJeanes: Also also I found that you can land on a very slight hill and keep sliding down slightly every time the brakes disengage and it gets in a bit of a loop until it finally settles, so I think we should also ignore very slow downward velocity

lua/entities/gmod_tardis/modules/sh_falling.lua
```diff
@@ -0,0 +1,191 @@
...
+                local newavel_z = math.Clamp(angv.z * 0.1,-300,300)
+                ph:SetAngleVelocityInstantaneous(Vector(newavel_x,newavel_y,newavel_z))
+            end
+        end
+
+        local pressing_down = IsValid(self.pilot) and TARDIS:IsBindDown(self.pilot,"flight-down")
+        local vertbrakes = self:GetData("vertbrakes")
+        local stopped = (vell < 0.1)
+
+        if pressing_down and not vertbrakes and vell > 5 then
```
AmyJeanes: The velocity tolerance seems a bit too low here, you can have the TARDIS flat on it's face with the slightest amount of movement and holding CTRL will make it stand up. Maybe it should be downwards velocity and a bit higher or do we intend for sliding on a flat surface to be able to make it vertical?

lua/entities/gmod_tardis/modules/sh_falling.lua
```diff
@@ -0,0 +1,191 @@
...
+            local ext=ply:GetTardisData("exterior")
+            e = ext or ply
+        end
+
+        if ply:GetTardisData("exterior")==self and e==self.thpprop and ply:GetTardisData("outside") then
+            self.wind_sound:ChangePitch(95+p,0.1)
+        else
+            local pos = e:GetPos()
+            local spos = self:GetPos()
+            local doppler = (pos:Distance(spos + e:GetVelocity()) - pos:Distance(spos + self:GetVelocity())) / 200
```
AmyJeanes: There still appears to be a popping sound when the sound first starts, you can clearly hear the 1 second delay it makes from it

AmyJeanes (changes requested):
This looks really great, fantastic work!

I've got a couple of number tweaking I'd like and a couple of maybe bugs but overall looking and working super well!

lua/entities/gmod_tardis/modules/sh_destination.lua
parar020100: Random destination can be found in the sky or on the ground
That's why I alternate them
I may just leave the grounded version here if you want 
(or make a condition for the sky pos like `shift`/`boost` pressed with G - what do u think?)

lua/entities/gmod_tardis/modules/sh_falling.lua
parar020100: we have a "TARDIS flip challenge" now, lol

What alternative behaviour do you propose for landing like that?

I definitely don't want the TARDIS bouncing around again, that never looks realistic, it aint a rubber ball. Judging by the first TARDIS scene of "deep breath", the TARDIS can instantly stabilize itself vertically after a very horizontal fall, as well as consuming the impact from a really high one ("NOTD"). 

Currently, the fall is not always vertical (as it used to be in the first version), but when it is vertical, the landing happens, and the drum makes it clear that some sort of physical brake is in place. So even though it might not be realistic with props, imo it fits the behaviour a TARDIS should have

parar020100: I'll see what I can do about sliding down

lua/entities/gmod_tardis/modules/sh_destination.lua
AmyJeanes: I think a condition would make sense yes!

lua/entities/gmod_tardis/modules/sh_falling.lua
AmyJeanes: I think as long as the horizontal velocity isn't stupidly high and the angular velocity is also not high then we can continue to let it do the landing. It should be able to stop from a small spin or small horizontal velocity I agree but too much is a jarring experience

lua/entities/gmod_tardis/modules/sh_destination.lua
parar020100: Made it always stick to the ground now

lua/entities/gmod_tardis/modules/sh_destination.lua
parar020100: aee08d3b9afe0165cf4b2aefa4426c9a835fa4b7
will ignore itself, but not other TARDISes

lua/entities/gmod_tardis/modules/sh_destination.lua
parar020100: c206c2b812d5ccd92f8922c939c71fbdf3bc505d
45b84a5536488bd6a5decf61fc479fc782cf91bb
I did the best I could

Plus, allowed SHIFT+R for another approach (same as right click), it's gonna be a rare use-case but could come in handy so no reason not to have it

lua/entities/gmod_tardis/modules/sh_falling.lua
parar020100: #include <discord_discussion.h>

e91a74397c27d5b0cc8d5ba5d93c2cb3637f6940

Sliding should be fixed in 0f10c6662c3cc5218eb2182547cb05b1d2cdcad9

lua/entities/gmod_tardis/modules/sh_falling.lua
parar020100: 0f10c6662c3cc5218eb2182547cb05b1d2cdcad9

lua/entities/gmod_tardis/modules/sh_falling.lua
parar020100: 93227df414d7ca11a26aeac9436f037ec7085988
c0909a95dd8ed0a23121336abf97291616ee2303

AmyJeanes (approved):

lua/entities/gmod_tardis/modules/sh_falling.lua
AmyJeanes: Removed and going to be resolved in a future PR

PR #962, opened 2023-10-22
A fix for dupes, saves, TARDISes without owner and more

### LINKED ISSUE #542: TARDIS doesn't work in world saves and dupes
Not sure whether this is fixable, but I'll leave it here

### LINKED ISSUE #949: Make the TARDIS owner-independent
- The TARDIS should be able to function without the creator being in the game
- The TARDIS should know its creator after they reconnect to the server

We should probably store the creator's nickname / steam id, and when one's missing, try to find them
The client settings should be stored / backed up at the entity so that when the creator is missing, it can continue existing with the same ones

### LINKED ISSUE #956: Keep chameleon circuit during redecoration

### LINKED ISSUE #957: Move door bodygroup related code to default addon's hooks
Complaint from VT:

Image

This code is creating some issues with the models from other extensions
Should be legacy only

### LINKED ISSUE #958: Random redecoration keeps selecting the same interior
From Discord

> Crimson Thunder
>  — 03/04/2023 21:22
> For some reason whenever you use redecoration without selecting an interior it just redecorates to the tardis you first spawn or the default. This happens to all extensions and doesn't seem to be because of a conflicting addon

### LINKED ISSUE #959: Allow chameleon circuit with no interior
From Za'ursus in Discord:

> If you set the option to create without an interior in the settings before creating the tardis, then this tardis will not be able to change the appearance of the exterior

parar020100:
Closes #949,#542,#957,#956,#958,#959

### DIFF
--- a/lua/entities/gmod_tardis/init.lua
+++ b/lua/entities/gmod_tardis/init.lua
@@ -31,20 +31,15 @@ ENT:AddHook("CustomData", "metadata", function(self, customData)
     end
 end)
 
-function ENT:InitializeReal()
-    self.exterior = self
-    self.metadata=TARDIS:CreateInteriorMetadata(self.metadataID, self)
-    self.Model=self.metadata.Exterior.Model
-    self.Portal=self.metadata.Exterior.Portal
-    self.Fallback=self.metadata.Exterior.Fallback
-    self.BaseClass.Initialize(self)
-end
-
 function ENT:Initialize()
-    if not IsValid(self:GetCreator()) then
-        return
+    if IsValid(self:GetCreator()) then
+        self.exterior = self
+        self.metadata=TARDIS:CreateInteriorMetadata(self.metadataID, self)
+        self.Model=self.metadata.Exterior.Model
+        self.Portal=self.metadata.Exterior.Portal
+        self.Fallback=self.metadata.Exterior.Fallback
+        self.BaseClass.Initialize(self)
     end
-    self:InitializeReal()
 end
 
 function ENT:PhysicsCollide(colData, collider)
@@ -58,6 +53,6 @@ end
 duplicator.RegisterEntityClass("gmod_tardis", function(ply, data)
     local ent = duplicator.GenericDuplicatorFunction(ply, data)
     ent:SetCreator(ply)
-    ent:InitializeReal()
+    ent:Initialize()
     return ent
 end, "Data")
--- a/lua/entities/gmod_tardis/init.lua
+++ b/lua/entities/gmod_tardis/init.lua
@@ -31,12 +31,19 @@ ENT:AddHook("CustomData", "metadata", function(self, customData)
     end
 end)
 
-function ENT:Initialize()
+function ENT:InitializeReal()
     self.metadata=TARDIS:CreateInteriorMetadata(self.metadataID, self)
     self.Model=self.metadata.Exterior.Model
     self.Portal=self.metadata.Exterior.Portal
     self.Fallback=self.metadata.Exterior.Fallback
     self.BaseClass.Initialize(self)
+end
+
+function ENT:Initialize()
+    if not IsValid(self:GetCreator()) then
+        return
+    end
+    self:InitializeReal()
 end
 
 function ENT:PhysicsCollide(colData, collider)
@@ -46,3 +53,10 @@ function ENT:OnTakeDamage(dmginfo)
     if self:CallHook("ShouldTakeDamage",dmginfo)==false then return end
     self:CallHook("OnTakeDamage", dmginfo)
 end
+
+duplicator.RegisterEntityClass("gmod_tardis", function(ply, data)
+    local ent = duplicator.GenericDuplicatorFunction(ply, data)
+    ent:SetCreator(ply)
+    ent:InitializeReal()
+    return ent
+end, "Data")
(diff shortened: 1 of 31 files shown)


lua/entities/gmod_tardis/init.lua
```diff
@@ -31,18 +31,33 @@ ENT:AddHook("CustomData", "metadata", function(self, customData)
...
     if self:CallHook("ShouldTakeDamage",dmginfo)==false then return end
     self:CallHook("OnTakeDamage", dmginfo)
-end
\ No newline at end of file
+end
+
+duplicator.RegisterEntityClass("gmod_tardis", function(ply, data)
+    local ent = duplicator.GenericDuplicatorFunction(ply, data)
+    ent:SetCreator(ply)
+    ent:InitializeReal()
```
AmyJeanes: I think you can just add the `if not IsValid(self:GetCreator())` check to the main Initialize function and then just call `Initialize` again, I don't think we need a separate `Real` function

lua/entities/gmod_tardis/modules/sh_doors.lua
```diff
@@ -147,18 +135,32 @@ if SERVER then
...
-    ENT:AddHook("ToggleDoor", "extcollision",function(self,open)
-        local door = TARDIS:GetPart(self,"door")
-        if IsValid(door) then
-            local override = self:CallHook("DoorCollisionOverride")
-            if (override == nil and open) or override==false then
-                door:SetSolid(SOLID_NONE)
-            elseif override or override==nil then
-                door:SetSolid(SOLID_VPHYSICS)
-            end
+    ENT:AddHook("ToggleDoor", "doorcollision",function(self,open)
```
AmyJeanes, line 138: Should these changes from #932 be in here?

lua/entities/gmod_tardis_interior/init.lua
```diff
@@ -15,7 +15,50 @@ ENT:AddHook("PostPlayerInitialize", "senddata", function(self,ply)
...
+        self:CallHook("AllowInteriorPos",nil,saved_pos,td.mins,td.maxs)~=false
+    then
+        return saved_pos
+    end
+
+    return pos
+end)
+
+ENT:AddHook("ShouldRemoveProp", "restore", function(self,prop)
+    if prop:IsVehicle() then
```
AmyJeanes, line 50: What's the reason for the `IsVehicle` here? I presume related to the seats in some interiors, but can we detect this in a better way e.g. set a special variable on them?

lua/entities/gmod_tardis_interior/init.lua
```diff
@@ -15,7 +15,50 @@ ENT:AddHook("PostPlayerInitialize", "senddata", function(self,ply)
     self:SendData(ply)
 end)
 
+ENT:AddHook("PostInitialize", "save_interior_position", function(self,ply)
+    self:SetData("interior_saved_pos", self:GetPos())
+end)
+
+ENT:AddHook("SetupPosition", "restore_position", function(self,pos)
+    local saved_pos = self:GetData("interior_saved_pos")
```
AmyJeanes, line 24: What would happen if you duplicated it on a different map? Should there be safeguards against this?

lua/entities/gmod_tardis_interior/modules/sh_power.lua
```diff
@@ -20,6 +20,12 @@ if SERVER then
         self.exterior:SetPower(on)
     end
 
+    ENT:AddHook("PostInitialize","power-init", function(self)
+        if self:GetData("power_disabled_first") then
```
AmyJeanes, line 24: Is this needed in the interior as well? The same code exists in the exterior

lua/entities/gmod_tardis_interior/init.lua
AmyJeanes: Update: Explained in DMs, difficult to lock onto vehicle entities tried many diff workarounds - good enough!

AmyJeanes (changes requested):
This works super well amazingly, fantastic work on this. Just a few tiny little questions / things but nothing major. Brilliant stuff.

lua/entities/gmod_tardis/modules/sh_doors.lua
parar020100: Dupes had problems with door collisions, code from #932 could fix that so I just re-added it.
There's no wall detection code though, just this refactoring

lua/entities/gmod_tardis_interior/modules/sh_power.lua
parar020100: we need all the hooks (including clientside ones) to work on the interior to set it to power off state

I tried avoiding it, but since the interior usually spawns after exterior, its lights and texturesets got completely messed up without this one

AmyJeanes: Okay cool, just wanted to make sure it was intended

lua/entities/gmod_tardis/init.lua
parar020100: e29bb179a52822815531ad20f7da637d85c8fefb

lua/entities/gmod_tardis_interior/init.lua
parar020100: e29bb179a52822815531ad20f7da637d85c8fefb

AmyJeanes (approved):
Fantastic work, really well done!

PR #969, opened 2023-11-02
Flight tracking

### LINKED ISSUE #56: A native way of 'flight tracking' from inside the TARDIS
Maybe in the GUI? Legacy requires sonic or E2 to do it, so I want a native method.

### LINKED ISSUE #321: Flight tracking ability
Same as legacy, feature parity

Linked with #56 to further enhance this but this issue is just for parity.

AmyJeanes:
Adds the flight tracking feature to the TARDIS

How to use:
- As the pilot by holding X (by default) in third person view to select a nearby entity
- Using the sonic screwdriver by linking your TARDIS then holding alt+right-click on a nearby entity
  - Requires updated sonic screwdriver, see below
- Using wiremod E2 function `entity:tardisTrack(entity)`

Once enabled, the pilot can shift the relative target around the tracked entity using standard flight mode controls, including rotation when the spinning is off.

To disengage, fly far enough away manually, remove the tracked entity, teleport, disable flight mode, press X or use the sonic to target nothing e.g. the ground or call `entity:tardisTrack()`

Fixes #321
Fixes #56

Note: Requires Sonic Screwdriver release: https://github.com/MattJeanes/Sonic-Screwdriver/commit/a329277976c94ea435c918c89ceed7ffaca0805a


i18n/languages/en.json
```diff
@@ -39,6 +39,10 @@
     "Binds.Sections.Flight.Rotate.Description": "Hold this key while using left and right to rotate",
     "Binds.Sections.Flight.SpinDirection": "Spin Direction",
     "Binds.Sections.Flight.SpinDirection.Description": "Changes which way the TARDIS rotates while flying",
+    "Binds.Sections.Flight.Tracking": "Tracking",
+    "Binds.Sections.Flight.Tracking.Description": "Hold to enable tracking selection mode",
```
parar020100: Not sure about the wording here, I wouldn't understand it if I were 
Perhaps we should have smth like "Hold to select the entity to track / follow"?

i18n/languages/en.json
```diff
@@ -228,6 +232,20 @@
...
     "Controls.ToggleScreens.Tip": "Toggle Screen",
+    "Controls.Tracking": "Tracking",
+    "Controls.Tracking.DistanceFail": "Can't track entities that are too far away",
+    "Controls.Tracking.FlightFail": "Failed to to enable flight mode for tracking",
+    "Controls.Tracking.GenericFail": "Failed to track entity",
+    "Controls.Tracking.InteriorFail": "Can't track entities inside the TARDIS",
+    "Controls.Tracking.PilotWarning": "WARNING: Tracking is enabled, press '%s' to change target or disable tracking",
+    "Controls.Tracking.Rotation": "Tracking rotation",
+    "Controls.Tracking.SameTarget": "You are already tracking this entity",
+    "Controls.Tracking.SelfFail": "Can't track itself",
```
parar020100: I would extend it just a bit, probably "The TARDIS can't track itself"

AmyJeanes: https://github.com/MattJeanes/TARDIS/pull/969/commits/2c72beec1d2ad53220acf32989f31d755e0c3e72

i18n/languages/en.json
AmyJeanes: https://github.com/MattJeanes/TARDIS/pull/969/commits/2c72beec1d2ad53220acf32989f31d755e0c3e72

lua/entities/gmod_tardis/modules/sh_spin.lua
```diff
@@ -24,52 +25,54 @@ end
...
+        self:SetData("spin-dir", dir, true)
+        self:CallHook("SpinChanged", self:GetSpinDir())
     end
 
-    function ENT:ToggleSpin()
-        local current = self:GetData("spindir", -1)
-        local prev = self:GetData("spindir_prev", 0)
+    function ENT:SetSpin(on, dir)
+        self:SetData("spin", on, true)
+        if on and dir ~= nil then
```
parar020100: Perhaps we should allow changing dir even if spin if off?

Just like we do here:
image

lua/entities/gmod_tardis/modules/sh_tracking.lua
```diff
@@ -0,0 +1,497 @@
...
+    end,
+    key=KEY_T,
+    serveronly=true,
+    exterior=true
+})
+
+if SERVER then
+    function ENT:GetTracking()
+        return self:GetData("tracking-ent")
+    end
```
parar020100: I will most likely want that to exist on the client
There are many potential, yet not planned, use-cases

Seeing info on the screen about the tracked entity would be pretty dope

lua/entities/gmod_tardis/modules/sh_tracking.lua
```diff
@@ -0,0 +1,497 @@
...
+    function ENT:SetTracking(ent,ply)
+        local wasTrackingEnt = self:GetData("tracking-ent")
+        local wasTracking = IsValid(wasTrackingEnt)
+        local isTracking = IsValid(ent)
+        local valid = true
+        if isTracking then 
+            if ent.TardisPart or ent.TardisInterior or (ent:IsPlayer() and IsValid(TARDIS:GetInteriorEnt(ent))) then
+                if IsValid(ply) then
+                    TARDIS:ErrorMessage(ply, "Controls.Tracking.InteriorFail")
+                end
```
parar020100: we don't need to check IsValid(ply) since that check exists at sh_notifications.lua:15

lua/entities/gmod_tardis/modules/sh_tracking.lua
```diff
@@ -0,0 +1,497 @@
...
+                valid = false
+            elseif ent == self then
+                if IsValid(ply) then
```
parar020100: > we don't need to check IsValid(ply) since that check exists at sh_notifications.lua:15

Same here

lua/entities/gmod_tardis/modules/sh_tracking.lua
```diff
@@ -0,0 +1,497 @@
...
+                    TARDIS:ErrorMessage(ply, "Controls.Tracking.SelfFail")
+                end
+                valid = false
+            elseif ent:GetPos():Distance(self:GetPos()) > TRACKING_MAX_DISTANCE_SET then
+                if IsValid(ply) then
```
parar020100: > we don't need to check IsValid(ply) since that check exists at sh_notifications.lua:15

Same here

lua/entities/gmod_tardis/modules/sh_tracking.lua
```diff
@@ -0,0 +1,497 @@
...
+                    TARDIS:ErrorMessage(ply, "Controls.Tracking.DistanceFail")
+                end
+                valid = false
+            elseif self:CallHook("CanTrack", ent, ply) == false then
+                if IsValid(ply) then
```
parar020100: > we don't need to check IsValid(ply) since that check exists at sh_notifications.lua:15

Same here

lua/entities/gmod_tardis/modules/sh_tracking.lua
```diff
@@ -0,0 +1,497 @@
...
+            end
+            if not valid then
+                return false
+            end
+
+            local wasFlying = self:GetFlight()
+            if not wasFlying then
+                local success = self:SetFlight(true)
+                if not success then
+                    if IsValid(ply) then
```
parar020100: > we don't need to check IsValid(ply) since that check exists at sh_notifications.lua:15

Same here

lua/entities/gmod_tardis/modules/sh_tracking.lua
```diff
@@ -0,0 +1,497 @@
...
+                valid = false
+            elseif self:CallHook("CanTrack", ent, ply) == false then
+                if IsValid(ply) then
+                    TARDIS:ErrorMessage(ply, "Controls.Tracking.GenericFail")
+                end
+                valid = false
+            end
+            if not valid then
+                return false
+            end
```
parar020100: Why use `valid` when you can just `return false` instantly after the notifications?

lua/entities/gmod_tardis/modules/sh_tracking.lua
```diff
@@ -0,0 +1,497 @@
...
+                if wasTrackingEnt ~= ent then
+                    local name = ent.PrintName or (isfunction(ent.Name) and ent:Name()) or ent.Name or ent:GetModel() or ent:GetClass()
+                    if ent.GetCreator then
+                        local creator = ent:GetCreator()
+                        if IsValid(creator) then
+                            name = name .. " (" .. creator:Nick() .. ")"
+                        end
+                    end
+                    TARDIS:Message(ply, "%s %s", "Controls.Tracking.Target", name)
+                else
```
parar020100: nit: trailing space
```suggestion
                else
```

lua/entities/gmod_tardis/modules/sh_tracking.lua
```diff
@@ -0,0 +1,497 @@
...
+        return self:GetData("tracking-ent")
+    end
+
+    function ENT:SetTracking(ent,ply)
+        local wasTrackingEnt = self:GetData("tracking-ent")
+        local wasTracking = IsValid(wasTrackingEnt)
+        local isTracking = IsValid(ent)
+        local valid = true
+        if isTracking then 
+            if ent.TardisPart or ent.TardisInterior or (ent:IsPlayer() and IsValid(TARDIS:GetInteriorEnt(ent))) then
```
parar020100: `ent.TardisPart`

this causes some issues

if you're trying to track a TARDIS exterior you sometimes select the exterior door and the tracking doesn't start and says "can't track entities inside the TARDIS"

Perhaps we should add something like:
```lua
if ent.TardisPart and ent.ExteriorPart then ent = ent.exterior end
```

One of the fun use-cases is chasing a TARDIS (e.g. if it's broken, it is hard to get close enough to start tracking), and when the trace goes to the door instead of the TARDIS it just becomes annoying

parar020100: Also gotta check for whether the exterior part belongs to this TARDIS, in which case it should be ignored as it is right now, but with a different message (Controls.Tracking.SelfFail)

lua/entities/gmod_tardis/modules/sh_tracking.lua
```diff
@@ -0,0 +1,497 @@
...
+            if wasTracking and (not self:GetData("tracking-wasflight")) then
+                self:SetFlight(false)
+            end
+            self:SetData("tracking-wasflight", nil)
+            self:SetData("tracking-offset-pos", nil)
+            self:SetData("tracking-offset-yaw", nil)
+            if IsValid(self.trackingdebugprop) then
+                self.trackingdebugprop:Remove()
+            end
+        end
```
parar020100: Most of the code has `if isTracking` conditions on it, we check for it too many times

Perhaps it would be better to have everything that happens when `not isTracking` to the beginning of the function?

lua/entities/gmod_tardis/modules/sh_tracking.lua
```diff
@@ -0,0 +1,497 @@
...
+                end
+            end
+
+            if not wasTracking then
+                self:SetData("tracking-wasflight", wasFlying)
+            end
+        end
+        
+        self:SetData("tracking-ent",ent)
+        if IsValid(ent) and ent ~= wasTrackingEnt then
```
parar020100: we're not changing `ent` before, why not use isTracking here?

lua/entities/gmod_tardis/modules/sh_tracking.lua
```diff
@@ -0,0 +1,497 @@
...
+        self:SetData("tracking-offset-yaw", newOffsetYaw)
+    end
+
+    function ENT:ToggleTrackRotation()
+        self:SetTrackRotation(not self:GetTrackRotation())
+    end
+
+    ENT:AddHook("Initialize", "tracking", function(self)
+        self:SetTrackRotation(true)
+    end)
```
parar020100, line 206: Are you sure you want this enabled by default?
Considering the fact that most players will not know the switch key, I think it would be best to make it either entity-dependent or false

I'd presume that the most often tracking would be of another TARDIS
Another obvious choice is an NPC or another player

NPCs and TARDISes rotating make this one go from side to side rather violently

You'd probably only want that rotation for vehicles by default

lua/entities/gmod_tardis/modules/sh_tracking.lua
```diff
@@ -0,0 +1,497 @@
...
+        ext:SetData("tracking-ent",ent)
+        if not IsValid(ent) then
+            local size = 10
+            local col = Color(255,0,0)
+            render.DrawLine(pos,pos+(fw*size),col)
+            render.DrawLine(pos,pos+(bk*size),col)
+            render.DrawLine(pos,pos+(ri*size),col)
+            render.DrawLine(pos,pos+(le*size),col)
+        end
+    end)
```
parar020100: This almost completely duplicates the code from cl_tp_rendering.lua:11-38

It is better to just implement it there with an extra condition 
(just like I did with the destination prop)

lua/entities/gmod_tardis/modules/sh_tracking.lua
```diff
@@ -0,0 +1,497 @@
...
+
+    ENT:AddHook("FlightToggled", "tracking", function(self, on)
+        if not on then
+            self:SetTracking(nil, self:GetData("pilot"))
+        end
+    end)
+
+    local VECTOR_0_0_1 = Vector(0,0,1)
```
parar020100: Unimportant nit-pick: could probably just call it VECTOR_UP, would make the code easier to read

lua/entities/gmod_tardis/modules/sh_tracking.lua
parar020100: Another thing to consider: should we track entities welded / constrained to the current TARDIS?
That currently causes some weird flying around

lua/entities/gmod_tardis/modules/sh_tracking.lua
```diff
@@ -0,0 +1,497 @@
...
+        if name == "GetTracking" then
+            return self:GetTracking() or NULL
+        end
+    end)
+
+    ENT:AddHook("FlightToggled", "tracking", function(self, on)
+        if not on then
+            self:SetTracking(nil, self:GetData("pilot"))
+        end
+    end)
```
parar020100: Not particularly related to these lines
Should we perhaps do the same for physlock enable?
The distance condition will probably turn it off after a while, but do we need it to wait?

lua/entities/gmod_tardis/modules/sh_tracking.lua
```diff
@@ -0,0 +1,497 @@
...
+                offsetrforce = offsetrforce*TARDIS:GetSetting("boost-speed")
+            end
+
+            local adjustedOffset = Vector()
+            if fbinds.forward then
+                adjustedOffset:Add(fwd * offsetforce)
+            end
+            if fbinds.backward then
+                adjustedOffset:Add(fwd * -offsetforce)
+            end
```
parar020100, line 287: **BUG**: when tracking-rotation is off, forward-back and right-left controls sometimes swap or work at wrong angles. 
E.g. you press forward and go right

It happened to me while I was tracking a patrolling NPC switching directions
But I presume this was related simply to the yaw of the tracked entity

parar020100: UPD: it does indeed depend on the tracked entity's yaw

lua/entities/gmod_tardis/modules/sh_tracking.lua
```diff
@@ -0,0 +1,497 @@
...
+        local tpdiff = targetpredicted-pos
+        local tdist = tpdiff:Length()
+        local diffnorm = tpdiff:GetNormalized()
+        local force = math.Clamp((tdist * 0.05), 0, 15*TARDIS:GetSetting("boost-speed"))
+
+        ph:AddVelocity(diffnorm*force*phm)
+
+        local brake = len-tdist
+        local brakeClamped = math.Clamp(brake, 0, len)*0.9
+        ph:AddVelocity(velnorm*-brakeClamped)
```
parar020100: nit-pick:
```suggestion
        ph:AddVelocity(-velnorm*brakeClamped)
```

lua/entities/gmod_tardis/modules/sh_tracking.lua
```diff
@@ -0,0 +1,497 @@
...
+            local ri = self:GetRight()
+            local ang = self:GetAngles()
+
+            local targetang = yawoffset
+            if trackrotation then
+                targetang = targetang + ent:GetAngles().y
+            end
+            local angdiff = math.AngleDifference(targetang,ang.y)
+            ph:AddAngleVelocity(Vector(0,0,angdiff*phm))
+            ph:AddAngleVelocity(Vector(0,0,ph:GetAngleVelocity().z*-0.3*phm))
```
parar020100: nit-pick:
```suggestion
            ph:AddAngleVelocity(Vector(0,0,-ph:GetAngleVelocity().z*0.3*phm))
```

parar020100 (changes requested):
This is indeed an awesome feature, you've done an awesome job!
I actually now regret not knowing it in the legacy, and it was hard to review since I kept getting distracted while testing. Running experiments - that's what people will start doing now :)

There are several issues I found with it, and some small nit-picks; feel free to ignore the latter if you feel like it.

There are two more things I suggest we consider:
- [x] I think there should be a way to make the TARDIS follow the owner. That is not currently possible neither from the sonic, nor from the interface. It would, however, be a very popular feature.

- [x] The interaction with the newly added broken flight is rather curious. Both in the damaged and broken states, the TARDIS tracking continues, but it has "jumps" from side to side. That's actually rather fun, but do we want tracking to work when it's broken?

    - [ ] Random idea: perhaps in the broken state it could start tracking some other random nearby entity? Lol. I like trolling players

lua/entities/gmod_tardis/modules/sh_tracking.lua
AmyJeanes: https://github.com/MattJeanes/TARDIS/pull/969/commits/95f488928ae50fc1df4c7b08f5bd8484657c2a63

lua/entities/gmod_tardis/modules/sh_tracking.lua
AmyJeanes: https://github.com/MattJeanes/TARDIS/pull/969/commits/3cfba78974d82469dd97008a3b366cbb3605ddc0 + https://github.com/MattJeanes/TARDIS/pull/969/commits/e87c55689360020c41f91ba39b676e0f9925a2c0

lua/entities/gmod_tardis/modules/sh_tracking.lua
AmyJeanes: https://github.com/MattJeanes/TARDIS/pull/969/commits/c0961ad6060f3421d2d38c467ed3b5ebf776a429

lua/entities/gmod_tardis/modules/sh_tracking.lua
AmyJeanes: I thought so too initially and had it implemented as such but it got more complicated because I needed to only draw the crosshair when we weren't looking at an entity, and it needed to set the `tracking-ent` data as well and it got needlessly complicated for the sake of saving a couple lines after adding the conditions to handle it.

It seemed a lot simpler to just do it all here instead even if a couple lines were copied. If you like we could move the `render.DrawLine` bits into a TARDIS helper function but not sure it's really needed.

lua/entities/gmod_tardis/modules/sh_tracking.lua
AmyJeanes: What if we edit the hint for "Tracking enabled" to something like "Tracking enabled, press 'KEY' to toggle tracking target rotation" or similar?

I don't want to presume what people are going to most likely using this for or try and build up some list of things to turn it on or off for as that will cause inconsistent behaviour for the user, so education on how to toggle it makes sense to me.

lua/entities/gmod_tardis/modules/sh_tracking.lua
AmyJeanes: https://github.com/MattJeanes/TARDIS/pull/969/commits/c5d1f4a007db37bb5c8a71e49e2a8ab9561b1e31

lua/entities/gmod_tardis/modules/sh_tracking.lua
AmyJeanes: It used to make more sense in a previous iteration of the code to reduce duplication, but that's no longer required

https://github.com/MattJeanes/TARDIS/pull/969/commits/153239cc2d0c2b3e43d8359e73c32fd1278d505f

lua/entities/gmod_tardis/modules/sh_tracking.lua
AmyJeanes: https://github.com/MattJeanes/TARDIS/pull/969/commits/5b22f4263b0ffd22a65c51f7013232fc202a9594

lua/entities/gmod_tardis/modules/sh_tracking.lua
AmyJeanes: https://github.com/MattJeanes/TARDIS/pull/969/commits/5b22f4263b0ffd22a65c51f7013232fc202a9594

lua/entities/gmod_tardis/modules/sh_tracking.lua
AmyJeanes: https://github.com/MattJeanes/TARDIS/pull/969/commits/5b22f4263b0ffd22a65c51f7013232fc202a9594

lua/entities/gmod_tardis/modules/sh_tracking.lua
AmyJeanes: https://github.com/MattJeanes/TARDIS/pull/969/commits/5b22f4263b0ffd22a65c51f7013232fc202a9594

lua/entities/gmod_tardis/modules/sh_tracking.lua
AmyJeanes: https://github.com/MattJeanes/TARDIS/pull/969/commits/5b22f4263b0ffd22a65c51f7013232fc202a9594

lua/entities/gmod_tardis/modules/sh_spin.lua
AmyJeanes: Don't see why not I suppose

https://github.com/MattJeanes/TARDIS/pull/969/commits/88d627855679156562a989a54535a3815630e1b1

lua/entities/gmod_tardis/modules/sh_tracking.lua
AmyJeanes: It also now auto disables physlock if it's engaged when activating tracking, like how it enables flight

https://github.com/MattJeanes/TARDIS/pull/969/commits/6bea59f4a5d583fbfd2f63d7553911e575638ea3

lua/entities/gmod_tardis/modules/sh_tracking.lua
AmyJeanes: https://github.com/MattJeanes/TARDIS/pull/969/commits/6bea59f4a5d583fbfd2f63d7553911e575638ea3

image

AmyJeanes:
Added tracking of player by attempting to track the TARDIS itself from the sonic: https://github.com/MattJeanes/Sonic-Screwdriver/commit/8114c6354e0f5cef350e63021bb7e6bd39f108ed

lua/entities/gmod_tardis/modules/sh_tracking.lua
AmyJeanes: Now treats any exterior parts as the exterior for tracking, and doesn't allow tracking constrained entities to itself

https://github.com/MattJeanes/TARDIS/pull/969/commits/711f9bfd97550f194d9be704dcec166c1080875c

Also, sonic has been updated to handle tracking exterior parts in the same way to track the player for consistent behaviour

https://github.com/MattJeanes/Sonic-Screwdriver/commit/177e6e5e6ba0374f6db33d98a3ba34a775729bd7

lua/entities/gmod_tardis/modules/sh_tracking.lua
AmyJeanes: Done

https://github.com/MattJeanes/TARDIS/pull/969/commits/7f6800965156a2e84bfba24a6c2b8ec4b3dc7725

Had to duplicate this little bit to make it work but otherwise looks a lot better now

```lua
if wasTracking ~= isTracking then
    TARDIS:StatusMessage(ply, "Controls.Tracking.Status", isTracking)
end
```

lua/entities/gmod_tardis/modules/sh_tracking.lua
parar020100: I really don't like duplicating code like that, it may with a high chance cause confusion in the future

May I suggest b44a7d5dbc4cd3f47b9944d459f2e647b44606cc? (`tracking_tweak` branch)

lua/entities/gmod_tardis/modules/sh_tracking.lua
parar020100: The first thing is great
I am not sure about "auto-disabling physlock" part

Seems like making things too easy
If you want that, perhaps it could be a setting? 

Same situation as auto-close doors during teleport, I believe it takes away a part of gameplay
The TARDIS as we've always known it is supposed to make you press a bunch of levers and buttons, run around the console etc - not automate controls for you

lua/entities/gmod_tardis/modules/sh_tracking.lua
parar020100: Okay

lua/entities/gmod_tardis/modules/sh_tracking.lua
parar020100: Seems to be okay now, thanks!

lua/entities/gmod_tardis/modules/sh_tracking.lua
AmyJeanes: Happy to remove, just thought might make things easier - but do we want it to fail to engage tracking if physlock is enabled or let it enable but just not move until physlock is disabled?

lua/entities/gmod_tardis/modules/sh_tracking.lua
AmyJeanes: I've merged that with some additional fixes/tweaks on top

https://github.com/MattJeanes/TARDIS/pull/969/commits/60e978a1e8c598799f99a67fc8c94a9a57d8fd25

lua/entities/gmod_tardis/modules/sh_tracking.lua
AmyJeanes: Added

https://github.com/MattJeanes/TARDIS/pull/969/commits/5e7e592d20c82abb64ad7c17ca179a8bab3b5dad

lua/entities/gmod_tardis/modules/sh_tracking.lua
```diff
@@ -0,0 +1,510 @@
+-- Tracking
+
+local TRACKING_MAX_DISTANCE_SET = 1000
```
AmyJeanes, line 3: This should take into account the bounding box of the target so we can target very large props

lua/entities/gmod_tardis/modules/sh_tracking.lua
AmyJeanes, line 1: Handle suggestions from https://discord.com/channels/811697995421581322/1171318512560640040/1171318512560640040

> At the moment, a very cool tracking function is in development. I would like to make a couple of suggestions to improve it.
> 
> If the tardis "B" follows the tardis "A", then when the tardis "A" is dematerialized, the first one will follow it into the vortex and will materialize together with the tardis "A".
> It will be possible to cancel the tracking and materialize separately, or if tardis "A" wants tardis "B" to stop following her, you will need to press the randomizer.
> 
> With the help of a sonic screwdriver, it will be possible to summon the tardis to itself, and instead of materializing next to it, it will fly on its own.
> 
> The ability to set tracking when choosing a destination, so that the tardis materializing, for example, near a flying plane, follows immediately after it.

AmyJeanes: Note: This doesn't mean we're adding these just marking it here so I don't forget, they may also become issues to add later rather than part of this PR

lua/entities/gmod_tardis/modules/sh_tracking.lua
parar020100: I'm happy with it just failing
But that's not a strong preference

lua/entities/gmod_tardis/modules/sh_tracking.lua
parar020100: okay!

lua/entities/gmod_tardis/modules/sh_tracking.lua
AmyJeanes: https://github.com/MattJeanes/TARDIS/pull/969/commits/65525f55a4954fd62d934d9f9d782735378f162c

lua/entities/gmod_tardis/modules/sh_tracking.lua
AmyJeanes: https://github.com/MattJeanes/TARDIS/pull/969/commits/4abca165a3bc2241764f4910ce0d5f9bd4afacb6

lua/entities/gmod_tardis/modules/sh_tracking.lua
AmyJeanes: - https://github.com/MattJeanes/TARDIS/issues/975
- https://github.com/MattJeanes/TARDIS/issues/976

parar020100 (approved):

PR #971, opened 2023-11-04
Added new Artron Energy functions for e2

NovaAstral:
- Added tardisEnergy() which returns the current Artron Energy of the TARDIS.
- Added tardisMaxEnergy() which returns the current 'tardis2_artron_energy_max' setting.
- Added tardisRemoveEnergy(number) which removes 'number' amount of energy from the TARDIS it is used on. This can be used if one wants to either easily drain a tardis, or for the intended use case, add fuel cost to custom e2 functions.

Unsure if this last function should only work on the Owners TARDISes or not.


AmyJeanes (approved):
Looks good to me! I think prop protection cantool is enough for removing artron energy, but we can adjust later if we see abuse. Note that I made a few little tweaks which you may want to see for future reference :)

PR #973, opened 2023-11-06
Added e2 functions for the new shields

NovaAstral:
- Added tardisShields() which toggles the shield
- Added tardisShields(number) which sets the shield
- Added tardisShieldsOn() which returns if the shields are on
- Added tardisShieldsLevel() which returns the shield amount
- Added tardisShieldsPercent() which returns the shield amount as a percentage
- Added tardisShieldsMax() which returns the max amount of shields the TARDIS can have


AmyJeanes (approved):
I've made some small changes to the naming and functionality but overall nice work!

PR #1073, opened 2024-05-13
TARDIS state & optimisations to be included

parar020100:
- Subfolder for lighting modules & split files 
- A quick way to call clientside hooks with no return value (2c0d38f3c36512e90a5b98f929b30de59a7d0500)
- TARDIS state added for matproxies & other use-cases
- Lots of matproxy optimisations
- Better way of specifying framerate for TARDIS_State_Texture matproxies (*)
- More refactoring
- Default interior: made the "lamps" version use 1 lamp instead of 7 (using interior size)
- Default interior: fixed the proper interior size for the smaller version (so that it's actually smaller, as it wasn't set)

(*)
Old way:
```vmt
			textures {
				idle "models/molda/toyota_int/bulbs"
...
				travel "models/molda/toyota_int/bulbs_travel"
				demat "models/molda/toyota_int/bulbs_travel"
...
				travel_warning "models/molda/toyota_int/bulbs_travel_warning"
				demat_warning "models/molda/toyota_int/bulbs_travel_warning"
				mat_warning "models/molda/toyota_int/bulbs_travel_warning"
			}
			frameRates {
				idle 0
				handbrake 0
...
				off 0
...
				demat_warning 40
				mat_warning 40
			}
```

New way:
```vmt
				dead "off" // copies the "off" state
...
				off "models/molda/toyota_int/bulbs_off"
				idle "models/molda/toyota_int/bulbs"
				handbrake "idle"

				travel { "models/molda/toyota_int/bulbs_travel" 10 }
				takeoff { "models/molda/toyota_int/bulbs_travel" 55 }
				parking { "models/molda/toyota_int/bulbs_travel" 50 }
...
```


lua/tardis/interiors/templates/templates_default_interior.lua
```diff
@@ -244,6 +217,14 @@ TARDIS:AddInteriorTemplate("default_small_version", {
     },
 })
 
+TARDIS:AddInteriorTemplate("default_small_version_lamp_fix", {
+    Interior = {
+        Size = {
+            Max = Vector(484.983, 514.944, 800)
```
AmyJeanes, line 223: What is this for?

lua/matproxy/matproxy_tardis.lua
```diff
@@ -1,12 +1,26 @@
 matproxy.Add({
-    name = "TARDIS_State",
+    name = "TARDIS_State_Texture",
```
AmyJeanes, line 2: Do any TARDISes use this already outside of base?

parar020100: Not that I know of

lua/tardis/interiors/templates/templates_default_interior.lua
parar020100: `default_small_version` is a template that makes the size smaller
`default_lamps` increases the interior height to fit the lamp (we only have one lamp now)

when they are both used, the lamp doesn't fit the small interior size, but this template adds the fix

AmyJeanes (approved):
Ran some basic tests in-game, looks pretty good to me. Nice work!

PR #1082, opened 2024-09-04
Feed activator entity into the CanPlayerEnterDoor hook

ryanm2711:
Useful if developers would like to do checks on the player before allowing them access to the TARDIS. For example in an RP scenario, if they are the right role.


AmyJeanes (approved):
Nice!

PR #1112, opened 2025-07-24
Overhauled material proxies to fix some issues

FuzzyLeo:
theres been a few bugs with material proxies when it comes to them functioning on the main entities for the interior and exterior, i've gone in an overhauled both my own and the original power one to allow those to function on the main entities, as a side thing i also made it so if its used on a tardis or a prop that does not feature the material proxy functionality, it will default to white, this both removes some errors being thrown and also allows people to spawn in the props without the colours being weird

### DIFF
--- a/lua/matproxy/matproxy_tardis.lua
+++ b/lua/matproxy/matproxy_tardis.lua
@@ -147,7 +147,16 @@ matproxy.Add({
 
     bind = function(self, mat, ent)
         if not IsValid(ent) or not ent.TardisPart then return end
-        if not ent.interior then return end
+        if not ent.interior.metadata.Interior.MatProxy then
+            local col = Color(255, 255, 255)
+
+
+            col = Color(col.r, col.g, col.b):ToVector()
+
+
+
+            mat:SetVector( self.ResultTo, col);
+        return end
 
         local col = ent.interior.metadata.Interior.MatProxy.Color1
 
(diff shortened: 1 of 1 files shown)


lua/matproxy/matproxy_tardis.lua
```diff
@@ -90,19 +90,27 @@ local function matproxy_tardis_power_init(self, mat, values)
...
-        self.last_value = value
+        local value = mat:GetVector(var)
+
+        if var ~= self.last_var or value ~= self.last_value then
+            self.last_var = var
+            self.last_value = value
+            mat:SetVector(self.ResultTo, value)
+        end
+    else
+        local value = Vector(1, 1, 1)
```
AmyJeanes: You wanna use the fallbackcol defined a bit later down here too?

lua/matproxy/matproxy_tardis.lua
```diff
@@ -146,17 +157,19 @@ matproxy.Add({
...
-        mat:SetVector( self.ResultTo, col);
+        if not IsValid(ent) then return end
+        if ent.interior then
+            ent = ent.interior
+        end
+        if ent.exterior then
+            if ent.metadata.Interior.MatProxy then
+                local col = ent.metadata.Interior.MatProxy.Color1
+                col = Color(col.r, col.g, col.b):ToVector()
+                mat:SetVector( self.ResultTo, col)
```
AmyJeanes: Inconsistent bracket spacing (also in a few other places in this file too)

lua/matproxy/matproxy_tardis.lua
```diff
@@ -211,16 +228,25 @@ local function matproxy_tardis_warning_init(self, mat, values)
...
-        self.last_value = value
+        local value = mat:GetVector(var)
+
+        if var ~= self.last_var or value ~= self.last_value then
+            self.last_var = var
+            self.last_value = value
+            mat:SetVector(self.ResultTo, value)
+        end
+    else
+        local value = Vector(1, 1, 1)
```
AmyJeanes: Use fallback colour here too?

lua/matproxy/matproxy_tardis.lua
FuzzyLeo: this isn't a colour, this is just a vector, why does the power value need a vector? no idea, maybe it would work using the fallback colour, i just didn't think to use it since in this case its not asking for a colour directly, just power state

FuzzyLeo:
made the requested changes

AmyJeanes (approved):

PR #1113, opened 2025-07-26
Material proxy that returns the light level of the exterior

FuzzyLeo:
essentially this would be used for things like tinting the light of interior windows, its able to read from map lighting and player placed lights, unfortunately i couldn't figure out how to make it react to lamps too.

i also made it so it fades to the vortex colour on demat, if a vortex colour isn't set it'll just fade to black to match the appreance in the husbands of river song since that can apply to any tardis

this could also be used for other things, for example tinting an exterior envmap to match the world lighting, and thats just what i can think of off the top of my head, people don't have to use it but it has a lot of potential i think so it would be nice if this was part of the base addon for everyone

also, credit to DoctorHugoSource for the overall idea of the implementation (the code itself is my own)

image
image
image
image
image

### DIFF
--- a/lua/matproxy/matproxy_tardis.lua
+++ b/lua/matproxy/matproxy_tardis.lua
@@ -284,3 +284,25 @@ matproxy.Add({
     init = matproxy_tardis_HDR_init,
     bind = matproxy_tardis_HDR_bind,
 })
+
+matproxy.Add({
+    name = "TARDIS_ExteriorBaseLight",
+
+    init = function(self, mat, values)
+        self.ResultTo = values.resultvar
+    end,
+
+    bind = function(self, mat, ent)
+        if not IsValid(ent) then return end
+        if ent.interior then
+            ent = ent.interior
+        end
+        if ent.exterior then
+            ent = ent.exterior
+            local col = render.ComputeLighting((ent:GetPos()+Vector(0, 0, 1)),Vector(0, 0, 1) ) -- Gets the position just above the tardis origin just in case it spawns slightly in the ground
+            mat:SetVector(self.ResultTo, col)
+        else
+            mat:SetVector(self.ResultTo, fallbackcol);
+        end
+    end
+})
--- a/lua/matproxy/matproxy_tardis.lua
+++ b/lua/matproxy/matproxy_tardis.lua
@@ -302,7 +302,7 @@ matproxy.Add({
             end
             vortexcol = Color(vortexcol.r, vortexcol.g, vortexcol.b):ToVector()
             ent = ent.exterior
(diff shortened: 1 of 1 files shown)


FuzzyLeo:
did some tweaks as i noticed there was room for improvement, its now actually ready to be merged

lua/matproxy/matproxy_tardis.lua
```diff
@@ -110,8 +110,6 @@ local function matproxy_tardis_power_bind(self, mat, ent)
             self.last_value = value
             mat:SetVector(self.ResultTo, value)
         end
-    else
```
AmyJeanes, line 113: Can you explain the removal of the fallback for this and the warning mat proxy?

lua/matproxy/matproxy_tardis.lua
```diff
@@ -283,4 +279,38 @@ matproxy.Add({
...
+            if ent.metadata.Interior.MatProxy then
+                if ent.metadata.Interior.MatProxy.VortexColor then -- Making sure a vortex colour is set in the first place since people tend to re-use door on multiple tardises
+                    vortexcol = ent.metadata.Interior.MatProxy.VortexColor
+                end
+            end
+            vortexcol = (Color(vortexcol.r, vortexcol.g, vortexcol.b):ToVector()*2.5)
+            ent = ent.exterior
+            local col = render.ComputeLighting((ent:GetPos()+Vector(0, 0, 80)),ent:GetForward()) -- Gets the lighting from the perspective of the doors
+            if ent:GetData("teleport") or ent:GetData("vortex") then
+                local ExteriorAlpha = (ent:GetData("alpha",255)/255)
```
AmyJeanes: Local variables should be lowercase

lua/matproxy/matproxy_tardis.lua
FuzzyLeo: the fallback replaces the value set in the vmt, and for some reason when respawning the tardis it would use that colour instead of the one it should, removing this fixed the bug and caused no other issues

AmyJeanes: Is this also the case for the other fallbacks that are still there?

FuzzyLeo: no as they are a different kind of value, warning and power switches between two values defined in the vmts whereas the colour ones just grab it elsewhere, it functions as intended with the colour ones, im not exactly sure why the warning and power ones get overridden but nonetheless this fix solves it without causing errors

FuzzyLeo: the fallback is intended so when they're spawned as props it doesn't inherit the colour from spawned tardises, its not required at all which is why removing it is fine, i will go in and take a proper look another time to hopefully prevent it inheriting power and warning states but comparitively its not a priority and i would rather avoid accidently implementing bugs where there weren't any before because of a fun personal tweak

AmyJeanes (approved):

PR #1115, opened 2025-07-29
Base Lighting exterior matproxy

FuzzyLeo:
the one i added last time was specific to the windows and wouldn't work as a real exterior base lighting matproxy, so here i've added a proper one designed specifically for the exterior and possible interior stuff, i also renamed the old one to better match what it actually does

heres some visual examples of its use case

image
Yes this first one looks unchanged, this one is here more to emphasise how it doesn't negatively effect how it looks in neutral lighting

image

image

image

essentially, this allows people to fix a long running issue with exterior tardis materials using envmaps causing them to look really wrong in anything except neutral lighting, especially in the dark as it would cause them to essentially glow, this is even an issue on the envmap based default exterior, and ESPECIALLY so on the included tt capsules which is most visible in the dark room on gm_construct

heres all some default capsules this issue currently effects, i've modified their vmts in this PR to fix them also
image


FuzzyLeo:
bump

Artixc:
I approve of this change, that env map looks way better. Ill shoot it up the pipeline

AmyJeanes (approved):

PR #1118, opened 2025-09-26
made some new transparency stuff opt-in for compatibility reasons

FuzzyLeo:
new transparency stuff can break tardises unfortunately, but this restores the original transparency fix with you having to opt-in to only using the new stuff, i think this is the best compromise unfortunately


FuzzyLeo:
made requested changes to make it opt-in per part via a new flag instead of changing the behaivour of the original transparency fix

AmyJeanes (approved):

PR #1122, opened 2025-10-30
made portal models compatible with chameleon circuit

FuzzyLeo:
just a minor fix as the chameleon circuit code didn't actually know to look for the new model parameters

### DIFF
--- a/lua/entities/gmod_tardis/modules/sh_chameleon.lua
+++ b/lua/entities/gmod_tardis/modules/sh_chameleon.lua
@@ -252,6 +252,9 @@ function ENT:ChangeExterior(id, animate, ply, retry)
             extportal:SetHeight(portal_md.height)
             extportal:SetThickness(portal_md.thickness or 0)
             extportal:SetInverted(portal_md.inverted)
+            extportal:SetModel(portal_md.model)
+            extportal:SetModelPos(portal_md.model_offset.pos)
+            extportal:SetModelAng(portal_md.model_offset.ang)
             extportal:SetParent(self)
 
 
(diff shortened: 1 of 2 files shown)


FuzzyLeo:
ok hold on this is not at all ready to be merged i fucked up

FuzzyLeo:
ok i've fixed it now, it can be merged

lua/entities/gmod_tardis/modules/sh_chameleon.lua
```diff
@@ -252,6 +252,13 @@ function ENT:ChangeExterior(id, animate, ply, retry)
             extportal:SetHeight(portal_md.height)
             extportal:SetThickness(portal_md.thickness or 0)
             extportal:SetInverted(portal_md.inverted)
+            if portal_md.model then
+                extportal:SetModel(portal_md.model)
```
AmyJeanes: Can this get unset again if switching back to one with no model? What is the behaviour here? Ditto with model offset

FuzzyLeo: haven't checked this, will check now

AmyJeanes: I'm happy for you to put this into the base interior metadata so it just always sets them rather than use custom logic here, agree that's a better approach.

Not sure on the models/error.mdl thing but it'll work for now. May change it later.

FuzzyLeo: done!

AmyJeanes (approved):

TimeLordian01:
Wait, so can I define a custom model shape with just an .mdl file now?

FuzzyLeo:
its only for the visuals, you still have to set up the position and bounds n stuff, but yes!  this was mainly to remove the shape limitations to allow things like circular portals

TimeLordian01:
Sick! So what would I need to put into an exterior file to define the model path?

FuzzyLeo:
just type add model = "model/path.mdl" somewhere in the portals table, same place as the "pos =" and other stuff

PR #1130, opened 2026-01-07
Offset parameter for portals

FuzzyLeo:
this is a small but useful addition to the parameters you can use with these portals, the idea is that templates can modify the offset position of the door allowing for more modular templates with doors, proves very useful for tardises with multiple exteriors that need matching interior doors, having essentially two separate position and angle parameters that get added together is overall just something thats handy i think and personally i'd get a lot of use out of it


AmyJeanes:
@FuzzyLeo did you mean to close this?

